PR #24, opened 2017-02-23
Body Encoding

sbebrys:
Support custom body encoding instead default utf8


StefH:
Some quick remarks:

Can you please only use `var` for complex types ? I like to keep string, bool and byte[] for readability.

And is it usefull to expose BodyEncoding also in the admin interface?

PR #32, opened 2017-06-10
[Feature] Add support for client certificate password and test with real services that require client certificate auth

phillee007:
Add support for client certificate password and test with real services that require client certificates. Also update so when proxying, the path and query will be passed to the proxied api.

Please have a look and let me know what you think. I ran a test for it locally, but it's a bit hard to checkin a runnable unit test as it would require a real service (which I have in my environment) plus a valid pfx with password

### DIFF
--- a/src/WireMock.Net/Admin/Mappings/ResponseModel.cs
+++ b/src/WireMock.Net/Admin/Mappings/ResponseModel.cs
@@ -84,5 +84,15 @@ namespace WireMock.Admin.Mappings
         /// </summary>
         /// <value>ProxyUrl</value>
         public string ProxyUrl { get; set; }
+
+        /// <summary>
+        /// The client X509Certificate2Filename to use.
+        /// </summary>
+        public string X509Certificate2Filename { get; set; }
+
+        /// <summary>
+        /// The X509Certificate2 password.
+        /// </summary>
+        public string X509Certificate2Password { get; set; }
     }
 }
--- a/src/WireMock.Net/Admin/Mappings/ResponseModel.cs
+++ b/src/WireMock.Net/Admin/Mappings/ResponseModel.cs
@@ -86,13 +86,8 @@ namespace WireMock.Admin.Mappings
         public string ProxyUrl { get; set; }
 
         /// <summary>
-        /// The client X509Certificate2Filename to use.
+        /// The client X509Certificate2 Thumbprint or SubjectName to use.
         /// </summary>
-        public string X509Certificate2Filename { get; set; }
-
-        /// <summary>
-        /// The X509Certificate2 password.
-        /// </summary>
-        public string X509Certificate2Password { get; set; }
+        public string X509Certificate2ThumbprintOrSubjectName { get; set; }
     }
 }
--- a/src/WireMock.Net/Http/CertificateUtil.cs
+++ b/src/WireMock.Net/Http/CertificateUtil.cs
@@ -22,14 +22,12 @@ namespace WireMock.Http
                     if (matchingCertificates.Count == 0)
                     {
                         // No certificates matched the search criteria.
-                        throw new Exception("no cert fount");
+                        throw new Exception($"No certificate found with Thumbprint or SubjectName '{thumbprintOrSubjectName}'");
                     }
                 }
-
                 // Use the first matching certificate.
                 return matchingCertificates[0];
             }
-
             finally
             {
                 if (certStore != null)
--- a/src/WireMock.Net/Http/CertificateUtil.cs
+++ b/src/WireMock.Net/Http/CertificateUtil.cs
@@ -1,0 +1,46 @@
+﻿using System;
+using System.Security.Cryptography.X509Certificates;
+
+namespace WireMock.Http
+{
+    internal static class CertificateUtil
+    {
+        public static X509Certificate2 GetCertificate(string thumbprintOrSubjectName)
+        {
+            X509Store certStore = new X509Store(StoreName.My, StoreLocation.LocalMachine);
+            try
+            {
+                //Certificate must be in the local machine store
+                certStore.Open(OpenFlags.ReadOnly);
+
+                //Attempt to find by thumbprint first
+                var matchingCertificates = certStore.Certificates.Find(X509FindType.FindByThumbprint, thumbprintOrSubjectName, false);
+                if (matchingCertificates.Count == 0)
+                {
+                    //Fallback to subject name
+                    matchingCertificates = certStore.Certificates.Find(X509FindType.FindBySubjectName, thumbprintOrSubjectName, false);
+                    if (matchingCertificates.Count == 0)
+                    {
+                        // No certificates matched the search criteria.
+                        throw new Exception("no cert fount");
+                    }
+                }
+
+                // Use the first matching certificate.
+                return matchingCertificates[0];
+            }
+
+            finally
+            {
+                if (certStore != null)
+                {
+#if NETSTANDARD || NET46
+                    certStore.Dispose();
+#else
+                    certStore.Close();
+#endif
+                }
+            }
+        }
+    }
+}
--- a/src/WireMock.Net/ResponseBuilders/Response.cs
+++ b/src/WireMock.Net/ResponseBuilders/Response.cs
@@ -37,7 +37,12 @@ namespace WireMock.ResponseBuilders
         /// <summary>
         /// The client X509Certificate2Filename to use.
         /// </summary>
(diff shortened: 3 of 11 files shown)


src/WireMock.Net/ResponseBuilders/Response.cs
```diff
@@ -272,6 +277,21 @@ public IResponseBuilder WithProxy(string proxyUrl, string clientX509Certificate2
...
         /// <summary>
+        /// With Proxy URL.
+        /// </summary>
+        /// <param name="proxyUrl">The proxy url.</param>
+        /// <param name="clientX509Certificate2Filename">The X509Certificate2 file to use for client authentication.</param>
+        /// /// <param name="clientX509Certificate2Password">The X509Certificate2 password.</param>
+        /// <returns>A <see cref="IResponseBuilder"/>.</returns>
+        public IResponseBuilder WithProxy(string proxyUrl, string clientX509Certificate2Filename, string clientX509Certificate2Password)
+        {
+            Check.NotEmpty(clientX509Certificate2Filename, nameof(clientX509Certificate2Password));
```
StefH: Fix this Check --> you check clientX509Certificate2Filename but use nameof(clientX509Certificate2Password)

And add a second check.

src/WireMock.Net/ResponseBuilders/Response.cs
```diff
@@ -272,6 +277,21 @@ public IResponseBuilder WithProxy(string proxyUrl, string clientX509Certificate2
         }
 
         /// <summary>
+        /// With Proxy URL.
+        /// </summary>
+        /// <param name="proxyUrl">The proxy url.</param>
+        /// <param name="clientX509Certificate2Filename">The X509Certificate2 file to use for client authentication.</param>
+        /// /// <param name="clientX509Certificate2Password">The X509Certificate2 password.</param>
```
StefH: remove ///

src/WireMock.Net/ResponseBuilders/Response.cs
```diff
@@ -285,7 +305,10 @@ public IResponseBuilder WithProxy(string proxyUrl, string clientX509Certificate2
 
             if (ProxyUrl != null)
             {
-                return await HttpClientHelper.SendAsync(requestMessage, ProxyUrl, X509Certificate2Filename);
+                var requestUri = new Uri(requestMessage.Url);
```
StefH: Why is this needed?

Why not just add the new parameter (X509Certificate2Password) to the method?

src/WireMock.Net/Server/FluentMockServer.Admin.cs
```diff
@@ -129,7 +129,11 @@ private void InitProxyAndRecord(ProxyAndRecordSettings settings)
 
         private async Task<ResponseMessage> ProxyAndRecordAsync(RequestMessage requestMessage, ProxyAndRecordSettings settings)
         {
-            var responseMessage = await HttpClientHelper.SendAsync(requestMessage, settings.Url);
+            var requestUri = new Uri(requestMessage.Url);
```
StefH: Same as above, please explain why this needed.

src/WireMock.Net/Admin/Mappings/ResponseModel.cs
```diff
@@ -84,5 +84,15 @@ public class ResponseModel
...
+
+        /// <summary>
+        /// The client X509Certificate2Filename to use.
+        /// </summary>
+        public string X509Certificate2Filename { get; set; }
+
+        /// <summary>
+        /// The X509Certificate2 password.
+        /// </summary>
+        public string X509Certificate2Password { get; set; }
```
StefH: Is it safe to return the password?

StefH:
I've added some remarks and questions. Can you please take a look?

phillee007:
Hi Stef,

Thanks for all your comments, they're very helpful and I see I missed a few things (inc. updating all tests!)
With regard to client certificates and passwords, the two main options I've seen in the past for reading a client cert to send with a request are:

1) Read it from a pfx file, and have to specify the password (as generally .pfx are stored with a password, especially if they're production certs we want to use when proxying a live service)
Pros:
- Can read from a NAS share or somewhere else, so tests can be run on any machine and read from a common location without having to do any importing of certs into their store
Cons:
- Have to send password for reading cert, and store in each request stub mapping and proxyandrecordsettings

2) Read it from the certificate store on the local machine
Pros:
- No need to store passwords in tests or test configuration
Cons:
- Have to configure machines on which tests will be run so that certificate is in the store

In this PR, I've done the code for option (1). However, I could update it to do option (2) instead if you think that's preferable? Thoughts?

Thanks,
Phil

StefH:
1. I'll do the review, thanks for taking this up. And provide feedback.
2. And I will have to read some more on certificates, to see what is possible.
3. You said you have implemented option 1, is is also possible to ALSO build option 2?

src/WireMock.Net/Http/CertificateUtil.cs
```diff
@@ -0,0 +1,46 @@
...
+                //Attempt to find by thumbprint first
+                var matchingCertificates = certStore.Certificates.Find(X509FindType.FindByThumbprint, thumbprintOrSubjectName, false);
+                if (matchingCertificates.Count == 0)
+                {
+                    //Fallback to subject name
+                    matchingCertificates = certStore.Certificates.Find(X509FindType.FindBySubjectName, thumbprintOrSubjectName, false);
+                    if (matchingCertificates.Count == 0)
+                    {
+                        // No certificates matched the search criteria.
+                        throw new Exception("no cert fount");
```
StefH: throw new Exception("No certificate found with Thumbprint or SubjectName '{0}'", thumbprintOrSubjectName);

src/WireMock.Net/Http/CertificateUtil.cs
```diff
@@ -0,0 +1,46 @@
...
+                    }
+                }
+
+                // Use the first matching certificate.
+                return matchingCertificates[0];
+            }
+
```
StefH: remove empty line

src/WireMock.Net/ResponseBuilders/Response.cs
```diff
@@ -264,34 +259,19 @@ public IResponseBuilder WithProxy(string proxyUrl)
         /// With Proxy URL.
         /// </summary>
         /// <param name="proxyUrl">The proxy url.</param>
-        /// <param name="clientX509Certificate2Filename">The X509Certificate2 file to use for client authentication.</param>
+        /// <param name="clientX509Certificate2ThumbprintOrSubjectName">The X509Certificate2 file to use for client authentication.</param>
         /// <returns>A <see cref="IResponseBuilder"/>.</returns>
-        public IResponseBuilder WithProxy(string proxyUrl, string clientX509Certificate2Filename)
+        public IResponseBuilder WithProxy(string proxyUrl, string clientX509Certificate2ThumbprintOrSubjectName)
```
StefH: You could add:
`[NotNull] string proxyUrl`
and 
`[NotNull] clientX509Certificate2ThumbprintOrSubjectName`

StefH:
I've added some small comments, can you take a look?

phillee007:
Yep all good. Having a bit of a nightmare with rebasing and trying to
revert some changes...

I don't think it's a good idea to include both options for file AND store -
at least not in the use cases I'm thinking of.

Might not be able to resume this for a few days, but will do my best.

Phil

StefH:
Yes rebasing and merging can be difficult to understand sometimes.

Maybe I can create an organization and add you as a member.

If we then work only from separate branches and review the code and only merge to master, that could also work...

phillee007:
If that works better for you I'm happy to do that. I could then create a
feature branch and reapply my changes to it instead of the fork. Sorry I
haven't worked with github forking and pull-requests a great deal so am
taking a while to get the hang of it...

phillee007:
Cool. I think some of the code in there is dev stuff that needs removing.
eg in the console app for the recorder example. Will check next time & do
another pr to fix

StefH:
I've just created an Github organization, and I will move the code from here to that organization.
After this is done, I will invite you.

StefH:
I sent invite to you for joining the organization.

Can you also tell me your full name ? So I can include this in the csproj/nuget.

As discussed ; can you please create a separate branch for new functionality, then we can review and if it's correct, merge back to master.

phillee007:
Cool, thanks! My full name is Phil Lee.

Will have a go at some of the unit tests for the code I changed next week,
plus do some more testing with proxying/record replay. Am thinking adding
an option to name files with url (minus nasty chars) + guid when in record
mode might be something i could add.

Cheers,
Phil

PR #35, opened 2017-06-16
Revert changes that were made by mistake in prior PR

phillee007:
Just changing the files back that I shouldn't have made in the first place. I wasn't expecting the PR to be merged so quickly :-)


StefH:
Ok.

PR #59, opened 2017-10-27
Add ability to provide multiple values for headers in response

Dreamescaper:
Add ability to specify several values for same header in response. In this PR it is not supported to define in json.

WithHeader and WithHeaders methods are backward-compatible. The only change, that could be breaking - public property Headers in ResponseMessage has different type now


StefH:
I'm updating the code to support the json.

Currently the code is working fine, so the response contains multiple headers:
```
MyHeader →application/json
MyHeader →application/json2
```

However, there are also some pages in internet which says that 
`MyHeader →application/json, MyHeader →application/json2`
or
`MyHeader →application/json; MyHeader →application/json2`

is also supported.

PR #60, opened 2017-10-28
Fix proxy headers handling

Dreamescaper:
Fix following issues with proxy (and tests added):
- content headers were not set in both request and response
- cookies were not set in request
- auto-redirect should not be performed when proxying requests
- if redirect response happened, and absolute Location host is the one that we're proxying to, we should change it to called one 
- creating new HttpClient for each request in not the best idea. I changed it so it would be created only once per ResponseBuilder.


StefH:
Thanks again!

PR #62, opened 2017-11-17
Add the Host, Protocol, Port and Origin to the Request message so they can be used in templating

alastairtree:
Hi

I needed to be able to template responses based on more granular information than was already available in the request object so have added some calculated fields and supporting test. Would you be interested in accepting it back?

- Add new fields for Host, Origin, protocol and Port during request message creation, derived from the Uri object
- Allow the new fields to be accessed in the templating engine
- Add tests for the new fields

### DIFF
--- a/appveyor.yml
+++ b/appveyor.yml
@@ -39,6 +39,6 @@ test_script:
   - nuget.exe install coveralls.net -ExcludeVersion
   - pip install codecov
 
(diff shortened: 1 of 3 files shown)


StefH:
Do you need a new NuGet for this PR?

alastairtree:
Great thank you. Yes please.

PR #63, opened 2017-11-17
Fix issue with concurrent logging

vladimir-fed:
Fixed issue with concurrent logging:
* added `ConcurrentObservableCollection` which is thread safe
* added test for checking logging with parallel requests
* added `GlobalExceptionMiddleware` that handles all exceptions that can happen in next middlewares
* changed response body from exception message to serialized exception when exception is thrown(it`s better to have whole exception fields than only exception message)


StefH:
Do you need a new NuGet for this PR?

vladimir-fed:
Yes, it will be great, thanks

PR #65, opened 2017-11-19
bug: Fix admin api client definition returning the wrong types

alastairtree:
Hello. Been testing the use of the admin api client with RestEase and noticed that the definition has the wrong types listed as being return on the get requests API meaning the restease client has null values and some data is missing. In addition the find requests API does not return LogEntryModel like the GetRequests API instead it returns LogEntry, which I have fixed so they are all the same.

- IFluentMockServerAdmin get, find and get all requests should return a log entry model
- Add tests for get and find using the rest ease api client


alastairtree:
Any chance of a release?

StefH:
Tonight

PR #67, opened 2017-11-20
bug: fix supporting the Patch method and logging the body

alastairtree:
If you need to mock an API designed to the JsonAPI specification you need to work with PATCH requests with a http body. This fix and tests should allow support of PATCH with body and ensure the correct information gets logged.

- Patch requests not configured to support Body
- Add a unit test covering a patch request and logging of results

### DIFF
--- a/src/WireMock.Net/Owin/OwinRequestMapper.cs
+++ b/src/WireMock.Net/Owin/OwinRequestMapper.cs
@@ -91,8 +91,9 @@ namespace WireMock.Owin
                 TRACE - Body not supported.
                 OPTIONS - Body supported but no semantics on usage (maybe in the future).
                 CONNECT - No defined body semantics
+                PATHC - Body suported
             */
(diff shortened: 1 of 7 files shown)


src/WireMock.Net/Owin/OwinRequestMapper.cs
```diff
@@ -91,8 +91,9 @@ private bool ParseBody(string method)
                 TRACE - Body not supported.
                 OPTIONS - Body supported but no semantics on usage (maybe in the future).
                 CONNECT - No defined body semantics
+                PATHC - Body suported
```
StefH: PATCH

alastairtree: Oops. Fixed

alastairtree:
Not sure but don't think this has been release? Could it be please?

StefH:
Done

Issue #73, opened 2017-12-19
Updated mapping is not being picked and responded with the response


raghavendrabankapur:
Start the FluentMockServer with below settings

```
string url = "http://localhost:1111/";
var server = FluentMockServer.Start(new FluentMockServerSettings
            {
                Urls = new[] { url },
                StartAdminInterface = true,
                ReadStaticMappings = true,
                ProxyAndRecordSettings = new ProxyAndRecordSettings
                {
                    Url = "http://www.google.com",
                    SaveMapping = true
                }
            });
```

Now access /__admin/mappings to get the default mappings in the console application example (having guid 11111110-a633-40e8-a244-5cb80bc0ab66). 
Update the pattern to /static/mapping/update
Once the mapping is updated, make a get call to localhost:1111/static/mapping/update, this will return 410 as response. 

Possible issue found is  when we try to find the matching mappings from the list of mappings we have. In WireMockMiddleWire.cs file in Invoke method we have a condition to to check if it's partial mappings or not. In the else loop, we are sorting the mapping object by priority and then return the IsPerfectMatch first or default value. Here we are finding another mapping with matches the above pattern with /* which is being registered when we add ProxyAndRecordSettings. So when we try to access localhost:1111/static/mapping/update it matches with /* and return 410 rather finding the other matches in the mapping.
If we can parse all the entries in the mappings object we should find the exact match for /static/mapping/update pattern also. Infact we can see in the mappings object and also the value of IsPerfectMatch being set to this result as true, but the first value which matches the pattern /* is being returned.

raghavendrabankapur:
![image](https://user-images.githubusercontent.com/5006495/34142614-2d4457f6-e4ae-11e7-8c76-95d094d6b80b.png)

The line of code which is commented in the screenshot was causing the problem. I understand this is used to record and proxy the request, but it was failing to map the request and respond accordingly

raghavendrabankapur:
![image](https://user-images.githubusercontent.com/5006495/34154715-84077130-e4dc-11e7-9b2e-a51358278498.png)

Above highlighted place is where we find the target mapping. In the else loop we order by priority and return the first element which will be matching the pattern /*

raghavendrabankapur:
@StefH I have found the solution for this. If I could create the pull request, then you can look into the fix and approve.

StefH:
You can make a PR and check if the unit-tests still work fine and maybe check if more tests are needed to test this scenario.

[...]

PR #74, opened 2017-12-20
Capturing the index of the existing mapping before removing and insert the updated mapping at the same index of the list

raghavendrabankapur:
This is to fix the issue #73 
We are inserting the updated mapping with the same guid in the same position rather than at the end of the list. This will keep the list state intact

### DIFF
--- a/src/WireMock.Net/Server/FluentMockServer.cs
+++ b/src/WireMock.Net/Server/FluentMockServer.cs
@@ -372,9 +372,32 @@ namespace WireMock.Server
         private void RegisterMapping(Mapping mapping)
         {
             // Check a mapping exists with the same GUID, if so, remove it first.
+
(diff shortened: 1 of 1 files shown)


raghavendrabankapur:
@StefH should I be adding a reviewer for this request? Because I cannot see an option but git say's reviewer required.

StefH:
I'm reviewer, and I'll review your change.

src/WireMock.Net/Server/FluentMockServer.cs
```diff
@@ -372,9 +372,32 @@ public IRespondWithAProvider Given(IRequestMatcher requestMatcher)
         private void RegisterMapping(Mapping mapping)
         {
             // Check a mapping exists with the same GUID, if so, remove it first.
```
StefH: // Check a mapping exists with the same GUID, if so, replace it.

PR #79, opened 2018-01-23
Fix missed content headers

vladimir-fed:
We need to pass the content headers even if the content is empty.
* fixed issue with passing content headers
* added test for checking the content header when the content is empty


StefH:
New version (1.0.2.13) will be uploaded to NuGet in a while.

vladimir-fed:
Thanks

Issue #94, opened 2018-02-27
Issue: Introduced dependency on log4net


alexangas:
Hello,

It looks like in v1.0.3 a dependency was introduced for log4net. I'm just wondering if it's possible to remove this please, or only include it for the Standalone server?

This doesn't significantly affect my project right now - at worst it is a mild annoyance as we use a different logging framework.

However for a project I worked on a year ago, this dependency would have made it impossible to use WireMock.Net. This is because for some reason, log4net is a candidate for abuse where developers create their own versions and embed it within their projects without changing the namespace. I've seen this twice in my career so far on significant projects such as NServiceBus (an old 2.x version but still in use), and an older version of Sitecore. It makes it impossible to include a reference to the official log4net binary.

Regardless, logging isn't required for the solution to operate, so it would be helpful if it was not included as a required dependency. 😃 

Can logging please be exposed as some sort of interface instead so developers can plug in their own logging code? It is simple to write.

Thanks for your consideration!

StefH:
I'll refactor the code to use logging interface which will probably look a lot like log4net or nlog.

StefH:
I've created a PR, and I want to add you as a reviewer, and I sent invite request for this repository.

PR #96, opened 2018-02-27
Replace log4net by custom logger (#94)

StefH:


StefH:
Can you approve this PR ?

(And you can always make a PR in the future to support implement more methods...)

PR #119, opened 2018-04-09
Fix for issue 118

raghavendrabankapur:
The code was not included to read the response from file


raghavendrabankapur:
Requesting a new pull request as this commit was void.

Issue #133, opened 2018-05-11
Issue: Wildcard matching a json body does not work?


alastairtree:
Hello @StefH, 

I have noticed that if I try to match request bodies containing json, I am restricted to using the JsonPathMatcher. Sometimes it is quicker and easier to use wildcard or regex matchers (say) but these matchers just get ignored if the content type header is set as application/json. This is not exactly intuitive - that a request header affects which of the matchers are applied or silently ignored. I assume this was a deliberate decision? Is there a reason you do not want to support the other matcher types on json? (and I assume the same is in place with xml.)

Sample failing test:

        [Fact]
        public async Task FluentMockServer_Should_respond_to_wildcard_matching_even_when_sent_as_json()
        {
            // Assign
            _server = FluentMockServer.Start();

            _server
                .Given(Request.Create().WithPath("/foo").WithBody(new WildcardMatcher("*Hello server*")))
                .RespondWith(Response.Create().WithBody(@"{ ""message"" : ""Hello client"" }"));

            // Act
            var content = new StringContent(@"{ ""message"" : ""Hello server"" }", Encoding.UTF8, "application/json");
            var response = await new HttpClient().PostAsync("http://localhost:" + _server.Ports[0] + "/foo", content);

            // Assert
            var responseString = await response.Content.ReadAsStringAsync();
            Check.That(responseString).Contains("Hello client");
        }

StefH:
It was not a 100% deliberate decision, this seemed logical at the time I built this piece of code.

I'll see if I can extend the matching logic to support this scenario, and not break others ;-)

StefH:
@alastairtree 
Please do a code-review on my `stef_133` branch.

See specifically the changes in `RequestMessageBodyMatcher.cs`.

alastairtree:
@StefH I have reviewed and added a couple more tests. I was wondering if it would be tidier to put the string version of the json in the BodyAsString property rather than adding a third?

StefH:
Thanks for tour tests, and `[Theory]` is cool option which I should more in my unit-tests in general.

If I remove the `` property, I think I need to switch the code into this:

``` c#
// Check if the matcher is a IStringMatcher
if (Matcher is IStringMatcher stringMatcher)
{
	// If the  body is a JSON object, try to use the Body to match.
	if (requestMessage.BodyAsJson != null && requestMessage.Body != null)
	{
		return stringMatcher.IsMatch(requestMessage.Body);
	}
	
[...]

PR #135, opened 2018-05-13
Merge into the stef_negate_matcher branch (solves issue #133)

StefH:


src/WireMock.Net/Matchers/Request/RequestMessageBodyMatcher.cs
```diff
@@ -105,27 +105,38 @@ public double GetMatchingScore(RequestMessage requestMessage, RequestMatchResult
...
+                if (requestMessage.Body != null)
+                {
+                    return stringMatcher.IsMatch(requestMessage.Body);
+                }
+
+                // If the  body is a JSON object, try to use BodyAsStringOriginal to match.
+                if (requestMessage.BodyAsJson != null && requestMessage.BodyAsStringOriginal != null)
+                {
+                    return stringMatcher.IsMatch(requestMessage.BodyAsStringOriginal);
+                }
```
alastairtree: So if we try and use Body here and remove BodyAsStringOriginal we can get by without introducing another version of the body on the requestMessage.

src/WireMock.Net/Util/BodyParser.cs
```diff
@@ -58,6 +59,7 @@ internal static class BodyParser
             else if (contentTypeHeaderValue != null && contentTypeHeaderValue.StartsWith("application/json", StringComparison.OrdinalIgnoreCase))
             {
                 var stringData = await ReadStringAsync(stream);
+                data.BodyAsStringOriginal = stringData.Item1;
```
alastairtree: And then put Item1 into Body instead and remove BodyAsStringOriginal?

src/WireMock.Net/Matchers/Request/RequestMessageBodyMatcher.cs
StefH: Fixed

src/WireMock.Net/Util/BodyParser.cs
StefH: Fixed

StefH:
Review comments fixed, please double check.

Issue #225, opened 2018-11-06
Feature: Improve logging in example for WireMock as Windows Service


paulssn:
As discussed in #223, the logging in the example for WireMock as Windows Service should be improved, by providing a log4net-logger as well as a log4net configuration with appenders for Console, File and Windows-EventLog.

paulssn:
@StefH You can assign this issue to me. I'll submit a PR.

paulssn:
Resolved by #247.

PR #247, opened 2018-12-23
Issue 225 improve logging in example for wire mock as windows service

paulssn:
Solves #225


examples/WireMock.Net.Service/log4net.config
```diff
@@ -0,0 +1,38 @@
...
+      </layout>
+    </appender>
+    <appender name="EventLogAppender" type="log4net.Appender.EventLogAppender" >
+      <layout type="log4net.Layout.PatternLayout">
+        <conversionPattern value="%date [%thread] %-5level %logger{1} - %message%newline" />
+      </layout>
+    </appender>
+    <root>
+      <level value="DEBUG" />
+      <appender-ref ref="ConsoleAppender" />
```
StefH: The `ConsoleAppender` is always enabled, doesn't this create issues when running as a windows service?

StefH: OK. Probably the ConsoleAppender from log4net has some logic built in.

PR #259, opened 2019-03-26
Fix Response.Delay property serialization issue

psypilat:


psypilat:
Fixed the problem with saving response mapping (also didn't show proper value via API). The problem was because of [Milliseconds](https://docs.microsoft.com/en-us/dotnet/api/system.timespan.milliseconds?view=netframework-4.7.2) property, [TotalMilliseconds](https://docs.microsoft.com/en-us/dotnet/api/system.timespan.totalmilliseconds?view=netframework-4.7.2) property should be used instead.

StefH:
Thanks for the PR, however, I'll change the parameter in the interface methods back to int, to keep it compatible.

PR #261, opened 2019-03-28
Fix BodyAsJson transform bug in ResponseMessageTransformer

psypilat:
Fix ResponseMessageTransformer to not replace BodyAsJson in an original message with transformed results


psypilat:
Want to mention that it means that if a consumer uses BadyAsJson with any handlebars functionality (transformer) it will work properly the first time, but then it will return the same response body for all next requests instead of generating them with transformers.

StefH:
I think I understand your issue. However, can you maybe create a unit-test to test this behaviour.

PR #262, opened 2019-03-29
Add ProvideResponse_WithJsonBodyAndTransform test

psypilat:


psypilat:
Sorry, unfortunately, I don't have time to create a full test suite, but I added one to ResponseWithBodyTests which shows the problem, fails when run against current version and passes when run against version with #262 fix.

StefH:
@psypilat 👍 
Thanks a lot !

PR #266, opened 2019-04-08
[265] Add file upload to allow mocking of file operations

JackCreativeCrew:
I've added endpoints for file upload/download/deletion (PUT/GET/DELETE) to allow for testing using files. The code is pretty basic but fulfills my requirements at the moment and has test coverage of happy paths.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,3 +1,7 @@
+# 1.0.11.0 (30 March 2019)
+- [#261](https://github.com/WireMock-Net/WireMock.Net/pull/261) - Fix BodyAsJson transform bug in ResponseMessageTransformer contributed by [psypilat](https://github.com/psypilat)
+- [#262](https://github.com/WireMock-Net/WireMock.Net/pull/262) - Add ProvideResponse_WithJsonBodyAndTransform test contributed by [psypilat](https://github.com/psypilat)
+
 # 1.0.10.0 (27 March 2019)
 - [#260](https://github.com/WireMock-Net/WireMock.Net/pull/260) - Fix Response.Delay property serialization [bug] contributed by [StefH](https://github.com/StefH)
 - [#257](https://github.com/WireMock-Net/WireMock.Net/issues/257) - Doc: Update outdated  [question]
--- a/Directory.Build.props
+++ b/Directory.Build.props
@@ -4,7 +4,7 @@
   </PropertyGroup>
 
   <PropertyGroup>
-    <VersionPrefix>1.0.12</VersionPrefix>
+    <VersionPrefix>1.0.13</VersionPrefix>
   </PropertyGroup>
 
   <Choose>
--- a/Directory.Build.props
+++ b/Directory.Build.props
@@ -4,7 +4,7 @@
   </PropertyGroup>
 
   <PropertyGroup>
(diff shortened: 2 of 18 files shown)


StefH:
Thanks for this idea + implementation.

I'll refactor the code in this branch to match the solution logic.

StefH:
@JackCreativeCrew 
I did update the code, and also added a `PostFile`.

Can you check the latest version on MyGet: https://www.myget.org/feed/wiremock-net/package/nuget/WireMock.Net ?

JackCreativeCrew:
Minor change to create the __admin/mapping folder if it doesn't exist otherwise the file upload fails. I'd imagine this is a problem in other places too (scenarios etc)?

StefH:
Actually I was just thinking about that scenario.

Also your remark bout returning the path.
--> this does not make sense because I think I will only allow writing files in the __admin/mapping and no not allow sub-folders.
Because withe current code you can use:
`POST http://{{wm_hostname}}/__admin/files/x/1.cs` which will try to save the file in the __admin/mapping/x folder.
And this fails because the "x" folder does not exists.

And actually, I think this is also a security risk to allow sub-folders or different paths like `/../../system32.dll` (as an example...)

So I think I will add logic to the code to take the filename only and change _path_ into _filename_ in the interface

JackCreativeCrew:
Yeah, ok, I hadn't considered that.

Something else I've come across, the file is being written as everything in the multipart body which isn't just the file. I've stuffed around trying to get this to write just the file but as yet failed.
I get for instance 
```
----------------------------566196384670995956765212
Content-Disposition: form-data; name=""; filename="pdf_viewer_resources.pak"
Content-Type: application/octet-stream

```
at the head and 
```

----------------------------566196384670995956765212--

```
at the tail. At the moment I'm only using one file i.e. that file, is there any way to only write the first piece of content and just chop the rest?

StefH:
First I will fix the unit-tests
Second I will think about multipart...

StefH:
**Second**
When I just upload a single file using Postman, only the file-content is uploaded, nothing to see as multipart...

JackCreativeCrew:
I must be doing something wrong on my end then. Thanks for checking that out.

StefH:
If you have tested all scenarios in this branch, and are confident that this code is correct, I can merge this PR and create a new NuGet.

JackCreativeCrew:
> If you have tested all scenarios in this branch, and are confident that this code is correct, I can merge this PR and create a new NuGet.

I haven't had time today to review/check my stuff against this today sorry. I'll get back to you by the end of tomorrow Sydney time though.

JackCreativeCrew:
Ok I've added a HEAD endpoint as well and a test for both file found and not found. It requires no body as the HEAD request can't have a body so the structure is a bit different. The code is now all working for me and if you accept those changes, I'm ok with merging it.

StefH:
Head is a good idea.

I did some small updates on code + tests, and I added this call also to the `IFluentMockServerAdmin.cs`

JackCreativeCrew:
Yeah, I was using that for debugging which admin endpoints were available, must've not gotten deleted before the commit.

StefH:
I'll merge to master now.

PR #293, opened 2019-06-28
workaround for AppContext.BaseDirectory being null on some platforms

eli-darkly:
This addresses https://github.com/WireMock-Net/WireMock.Net/issues/292 which is known to affect Xamarin Android.

I'm not aware of any other platforms where this is a problem, but if there are any, the principle is the same: if AppContext.BaseDirectory is null, then WebHostBuilder won't work without setting ContentRoot. WireMock.Net wouldn't work at all under that condition before, so the workaround can't hurt - and the workaround is only used in that condition, so platforms where WireMock.Net _did_ work are not affected.

I don't know of any way to unit-test this change, since it's not possible to change AppContext.BaseDirectory dynamically in a test. However, I have tested manually in Xamarin Android and it solved the problem for me.

### DIFF
--- a/src/WireMock.Net/Owin/AspNetCoreSelfHost.cs
+++ b/src/WireMock.Net/Owin/AspNetCoreSelfHost.cs
@@ -56,7 +56,18 @@ namespace WireMock.Owin
 
         public Task StartAsync()
         {
-            _host = new WebHostBuilder()
+            var builder = new WebHostBuilder();
+
+            // Workaround for https://github.com/WireMock-Net/WireMock.Net/issues/292
+            // On some platforms, AppContext.BaseDirectory is null, which causes WebHostBuilder to fail if ContentRoot is not
+            // specified (even though we don't actually use that base path mechanism, since we have our own way of configuring
+            // a filesystem handler).
+            if (string.IsNullOrEmpty(AppContext.BaseDirectory))
+            {
+                builder.UseContentRoot(System.IO.Directory.GetCurrentDirectory());
+            }
+
+            _host = builder
                 .ConfigureServices(services =>
                 {
                     services.AddSingleton(_options.FileSystemHandler);


StefH:
Thanks for this PR.

Is it possible to also create xamarin unit tests? Or is that not supported?

eli-darkly:
> Is it possible to also create xamarin unit tests?

There's no single Xamarin runtime - there's Xamarin Android, Xamarin iOS, etc. - although Xamarin Android is the only one with this issue as far as I know. So you have to actually run the targeted OS, or at least an emulator for it. There is a way to run Xunit tests inside of an Android emulator, I've done it, but it's extremely cumbersome and slow and I don't know how to set it up in Azure.

eli-darkly:
(Here's what I mean by "cumbersome": https://gist.github.com/eli-darkly/d3d33aa1cea67a1ce7935e9cb6676e1b - I think it is probably possible to do the equivalent in Windows, which might even be a bit faster since the Android emulator is known to perform very badly on Linux.)

StefH:
Another idea : Can you also create a simple xamarin android exanple app which uses wiremock.net and place this in the examples folder?

eli-darkly:
> Can you also create a simple xamarin android exanple app which uses wiremock.net and place this in the examples folder?

@StefH I guess, although that wouldn't really be an example of how to use WireMock - the usage of the API would be exactly the same as all the other examples. The only point of it would be to prove that if you run the app, it doesn't crash. The other thing is that the scenario I'm talking about is when you're running _unit test_ code in Xamarin Android... which requires that you select a particular test framework and test runner app, since there is no built-in mechanism for doing this, so the example would be specific to (for instance) Xunit and the xunit.runner.devices package.

StefH:
Ok. Thanks for the explanation.

eli-darkly:
I'm happy to give it a try - I don't like to submit code that has no tests at all (although at least the existing tests should be enough to show that this change didn't break anything for apps that are _not_ using Xamarin Android) - just warning you that it would probably also involve a bit of refactoring of your existing examples.

StefH:
It's not needed.

Issue #301, opened 2019-07-22
Error thrown when calling mocked endpoint second time when using file-based response body


rwwilden:
When you use a setup as follows:

    server
      .Given(
          Request
              .Create()....)
      .RespondWith(
          Response
              .Create().....
              .WithBodyFromFile(fileName));

the mocked endpoint can only be called successfully the first time. Second time it will throw an error.

This is caused by the fact that the property `BodyData.BodyAsFile` is set to `null` right after the first successful response, causing an `ArgumentNullException` the second time. This happens in the [`Response`](https://github.com/WireMock-Net/WireMock.Net/blob/master/src/WireMock.Net/ResponseBuilders/Response.cs#L422) class.

StefH:
Dankjewel.

PR #302, opened 2019-07-22
Fixed bug 301 by not setting BodyAsFile to null after first use

rwwilden:
#301


StefH:
Thanks. I did add an extra unit-test to test this scenario.

Issue #287, opened 2019-06-19
Error with parameter that contains a "=" character


guerital:
I have created a small sample to report the error I'm facing with:

``` c#
using System.Net;
using NUnit.Framework;
using WireMock.Matchers;
using WireMock.RequestBuilders;
using WireMock.ResponseBuilders;
using WireMock.Server;
using WireMock.Settings;

namespace WireMockNUnit
{
    [TestFixture]
    class ReportError
    {
        #region Fixture

        private FluentMockServer _server;

        #endregion

        #region Setup & Teardown

        [SetUp]
        public void SetUp()
        {
            _server = FluentMockServer.Start(new FluentMockServerSettings
            {
                Urls = new[] {"http://localhost:12345/" },
                StartAdminInterface = true
            });
        }

        [TearDown]
        public void TearDown()
        {
            _server.Stop();
        }

        #endregion

        [TestCase("value")]
        [TestCase("va?ue")]
        [TestCase("val=e")]
        public void Test(string value)
        {
            _server
                .Given(Request
                    .Create()
                    .WithPath("/path")
                    .WithParam("key", MatchBehaviour.AcceptOnMatch, true, value)
                    .UsingGet())
                .RespondWith(Response
                    .Create()
                    .WithStatusCode(200)
                    .WithBody("You got it!"));

            var client = new WebClient();
            var content = client.DownloadString($"http://localhost:12345/path?key={value}");

            Assert.AreEqual(content, "You got it!");
        }
    }
}
```

Running this test that has three test case I have obtained the following result:
* case "value": No error
* case "va?ue": No error
* case "val=e": 
```
System.Net.WebException : The remote server returned an error: (404) Not Found.
   at System.Net.WebClient.DownloadDataInternal(Uri address, WebRequest& request)
   at System.Net.WebClient.DownloadString(Uri address)
   at WireMockNUnit.ReportError.Test(String value) in C:\Users\italo.guerrieri\Downloads\WireMockPoc\WireMockNUnit\Class1.cs:line 57
```

lobsteropteryx:
I've submitted a PR which I believe will fix this problem; in our case, we were trying to send a parameter with a double equals `==`, and we found that we could match properly on the same querystring with a single equals `=` in our test code.

For example, our production code would send `?key=value==something`, and our test code would match on `?key=value=something`.  Don't know if that helps, but thought it might be worth mentioning!

StefH:
@lobsteropteryx  Thank you.

[...]

PR #316, opened 2019-08-13
Handles case where parameter value contains ==

lobsteropteryx:
This is a fix for #287; querystring parameters that contain `==` in their value were not being properly parsed, i.e., `?key=value==something`.

The fix here limits splitting each querystring parameter to the first `=` in the string.

I don't see any contributing guidelines, so I'm happy to make changes if need be.  Thank you for a great tool!

### DIFF
--- a/src/WireMock.Net/Util/QueryStringParser.cs
+++ b/src/WireMock.Net/Util/QueryStringParser.cs
@@ -18,11 +18,6 @@ namespace WireMock.Util
 
             string[] JoinParts(string[] parts)
             {
-                if (parts.Length > 2)
-                {
-                    return new[] { string.Join("=", parts, 1, parts.Length - 1) };
-                }
-
                 if (parts.Length > 1)
                 {
                     return parts[1].Split(new[] { "," }, StringSplitOptions.RemoveEmptyEntries); // support "?key=1,2"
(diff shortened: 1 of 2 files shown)


StefH:
Thanks!

Did you also test this PR  ode in a real scenario?

lobsteropteryx:
Hi @StefH!

Yes, I've pulled the changes into a local project which is using WireMock.Net for both unit tests, and as a standalone process for acceptance tests.  I can't share that code, but would be happy to add a higher level test to the WireMock repo if that would help.  Do you have any suggestions?

StefH:
@lobsteropteryx 
No extra tests are needed.

Can you maybe share a mapping json file and a test request ? Then I can replay that in my local wiremock to be 100% sure.

lobsteropteryx:
Sure thing...I'll put them inline here.

A mapping like:

```json
{
    "Guid": "efd5a2e5-6575-4d81-a9bd-15e0bcfee0f8",
    "Request": {
      "Path": {
        "Matchers": [
          {
            "Name": "WildcardMatcher",
            "Pattern": "/test",
            "IgnoreCase": false
          }
        ]
      },
      "Methods": [
        "GET"
      ],
      "Params": [
        {
          "Name": "q",
          "Matchers": [
            {
              "Name": "ExactMatcher",
              "Pattern": "param==1"
            }
          ]
        }
      ]
    },
    "Response": {
      "StatusCode": 200,
      "BodyDestination": "SameAsSource",
      "Body": "{\"success\": true}", 
      "Headers": {
        "Content-Type": "application/json"
      }
    }
  }
```
Should work with a request like

```
curl "localhost:9091/test?q=param==1"
```

StefH:
Thanks.

PR #331, opened 2019-08-22
Fix: Collection was modified exception

theramis:
## Problem
We are using wiremock as part of our tests and we found that if we have tests running in parallel we sometimes get an error when calling the `FindLogEntries` method. 

**Error** 
```
System.InvalidOperationException: Collection was modified; enumeration operation may not execute.
```

## Potential Solution
This is my quick attempt at fixing the above issue. By convert the collection into a new list it shouldn't throw an error if the `LogEntries` collection is updated. 
From a quick google around the `ToList` method itself isn't thread safe either though. So its possible that while the `ToList` method is creating a copy, it fails due to the original collection getting changed.

Thoughts?
@StefH

### DIFF
--- a/src/WireMock.Net/Server/FluentMockServer.LogEntries.cs
+++ b/src/WireMock.Net/Server/FluentMockServer.LogEntries.cs
@@ -39,7 +39,7 @@ namespace WireMock.Server
         {
             var results = new Dictionary<LogEntry, RequestMatchResult>();
 
-            foreach (var log in _options.LogEntries)
+            foreach (var log in _options.LogEntries.ToList())
             {
                 var requestMatchResult = new RequestMatchResult();
                 foreach (var matcher in matchers)


src/WireMock.Net/Server/FluentMockServer.LogEntries.cs
```diff
@@ -39,7 +39,7 @@ public IEnumerable<LogEntry> FindLogEntries([NotNull] params IRequestMatcher[] m
         {
             var results = new Dictionary<LogEntry, RequestMatchResult>();
 
-            foreach (var log in _options.LogEntries)
+            foreach (var log in _options.LogEntries.ToList())
```
StefH: I think it's even better to just use the property `LogEntries` here (that property is already made safe...)


Also in the `DeleteLogEntry()` method, it's safer to use the `LogEntries` property.

StefH:
Thanks for this PR, can you take a look at my comments?

src/WireMock.Net/Server/FluentMockServer.LogEntries.cs
theramis: > I think it's even better to just use the property LogEntries here (that property is already made safe...)

Adding/Removing/Setting/Clearing is thread safe for the `LogEntries` property (due to https://github.com/WireMock-Net/WireMock.Net/blob/master/src/WireMock.Net/Util/ConcurrentObservableCollection.cs) but enumerating it is not. 

In my case we have thread 1 calling `FindLogEntries` 
and thread 2 calling a mocked http api. 

Since the `foreach` loop above does not hold a lock, the log entry for the call made by thread 2 is added to the `LogEntries` property. This causes the enumerator in the `foreach` loop to throw an exception. 

I hope that explanation makes sense?

theramis: > Also in the DeleteLogEntry() method, it's safer to use the LogEntries property

Yup in that method `LogEntries` is perfect, since the `Remove` method on `LogEntries` is thread safe.

gregoks:
hi guys,

I'm using version 1.0.30-ci11848 and still getting this error:

HttpStatusCode set to 500 System.InvalidOperationException: Collection was modified; enumeration operation may not execute. at System.Collections.Generic.List`1.Enumerator.MoveNextRare() at System.Linq.Enumerable.WhereEnumerableIterator`1.ToList() at System.Linq.Enumerable.ToList[TSource](IEnumerable`1 source) at WireMock.Owin.WireMockMiddleware.LogRequest(LogEntry entry, Boolean addRequest) at WireMock.Owin.WireMockMiddleware.InvokeInternal(HttpContext ctx) at WireMock.Owin.GlobalExceptionMiddleware.InvokeInternal(HttpContext ctx)

Is it the same as in this bug?

StefH:
@gregoks A new fix is required I think.

StefH:
@gregoks Can you please try: 
WireMock.Net.1.0.31-ci-11865

PR #334, opened 2019-08-26
Fix issues with Proxy mode and Binary Request Bodies

andi0b:
We used Wiremocks.NET in Proxy mode to sometimes test against the real API and sometimes just mock it. This stopped working when we started POSTing JPEGs. The body was altered because WireMocks.NET started to parse the binary content as UTF8 and converted it back. This causes issues with binary contents, you can see it because the length changes.

This pull request contains:
- Test to show this issue
- Fix for this issue (by replacing Encoding.UTF8 with another UTF8 decoder which throws Exceptions on unmappable characters, so the code can fallback to binary parsing)
- Another fix in RequestMessageBodyMatcher I discovered because my test didn't work correctly. (Enable binary matching also if the content is detected as string or JSON).


src/WireMock.Net/Matchers/Request/RequestMessageBodyMatcher.cs
```diff
@@ -142,20 +142,20 @@ private double CalculateMatchScore(RequestMessage requestMessage)
             {
                 return Matchers.Max(matcher => CalculateMatchScore(requestMessage, matcher));
             }
-
-            if (Func != null)
+            
+            if (Func != null && requestMessage?.BodyData?.BodyAsString != null)
```
StefH: With this extra `&&`, the Func is not executed anymore when the BodyAsString is null.
It could be that you actually want act on this. So better to keep original if-statement, and make sure that the function-call is null-safe. See comment below.

src/WireMock.Net/Matchers/Request/RequestMessageBodyMatcher.cs
```diff
@@ -142,20 +142,20 @@ private double CalculateMatchScore(RequestMessage requestMessage)
...
             {
-                return MatchScores.ToScore(requestMessage?.BodyData?.DetectedBodyType == BodyType.String && Func(requestMessage.BodyData.BodyAsString));
+                return MatchScores.ToScore(Func(requestMessage.BodyData.BodyAsString));
```
StefH: Use safe null-check here: `requestMessage?.BodyData?.BodyAsString`

test/WireMock.Net.Tests/FluentMockServerTests.Proxy.cs
```diff
@@ -308,6 +309,43 @@ public async Task FluentMockServer_Proxy_Should_preserve_cookie_header_in_proxie
             Check.That(receivedRequest.Cookies).ContainsPair("name", "value");
         }
 
+        /// <summary>
```
StefH: can you please also add a more low-level unit-test to test just this class with the changed logic for UTF8 ?
See https://github.com/WireMock-Net/WireMock.Net/blob/master/test/WireMock.Net.Tests/Util/BodyParserTests.cs for the tests.

andi0b:
@StefH
> Thanks for the PR.
> 
> Can you please take a look at my comments?

I agree with your comments, I will do it as suggested.

andi0b:
@StefH Please review my changes. I added some test coverage, but I have to admit I don't fully understand the multipart thing in BodyParser.

I'm not used to a pull request workflow, is there anything else I should do? Merge sth?

StefH:
Hello @andi0b,

1. It's better to make changes in a separate branch, and then make a PR to merge to my master. In this way, I can also get your branch and make adjustments if needed.

2. There is a merge conflict in BodyParser.cs
To solve this, you must get latest master from my project en solve merge conflicts.

3. You cannot merge. Once the code is ok, and all my merge comments are resolved, then I can merge this to master.

StefH:
4. Please check you whitespace setting in Visual Studio.
I use 4 spaces, and never tabs.

If you use different, a lot of code changes will be displayed.

andi0b:
>     1. Please check you whitespace setting in Visual Studio.
>        I use 4 spaces, and never tabs.

Yeah, actually I use Rider with auto detect of code style (as there is no .editorconfig), and it's using 4 spaces, no idea what happened there... maybe it tried to outsmart all of us.

But there is also something weird with this change: https://github.com/WireMock-Net/WireMock.Net/commit/c0a43ed20475778ca0661a18c3e8d00305382c54#diff-78e158879161b0265ca8030bed233919

andi0b:
I don't really know what to do about the failing test `FluentMockServer_Should_exclude_body_for_methods_where_body_is_definitely_disallowed`. It used to pass because of the bug I fixed with the matching.

So the first part never matched because the content isn't json:
```cs
            // Assign
            string content = "hello";
            var server = FluentMockServer.Start();

            server
                .Given(Request.Create().WithBody(b => true))
                .AtPriority(0)
                .RespondWith(Response.Create().WithStatusCode(400));
```

StefH:
But there is also something weird with this change: c0a43ed#diff-78e158879161b0265ca8030bed233919

--> Yes, I think that the previous fix in that file (done by someone else) also used a different space/tabs setting.

PR #346, opened 2019-09-16
Fix recorded requests skipped by request logger

vitaliydavydiak:
When proxy is enabled the recorded requests are mistaken (IMO) for admin interface requests and skipped in LogEntries. Actually none of requests logged in proxy mode. It is valuable to monitor that while recording.

### DIFF
--- a/src/WireMock.Net/Mapping.cs
+++ b/src/WireMock.Net/Mapping.cs
@@ -46,7 +46,10 @@ namespace WireMock
         public bool IsStartState => Scenario == null || Scenario != null && NextState != null && ExecutionConditionState == null;
 
         /// <inheritdoc cref="IMapping.IsAdminInterface" />
-        public bool IsAdminInterface => Provider is DynamicResponseProvider || Provider is DynamicAsyncResponseProvider || Provider is ProxyAsyncResponseProvider;
+        public bool IsAdminInterface => Provider is DynamicResponseProvider || Provider is DynamicAsyncResponseProvider;
+
+        /// <inheritdoc cref="IMapping.IsRecordedByProxy" />
+        public bool IsRecordedByProxy => Provider is ProxyAsyncResponseProvider;
 
         /// <summary>
         /// Initializes a new instance of the <see cref="Mapping"/> class.
--- a/src/WireMock.Net/Mapping.cs
+++ b/src/WireMock.Net/Mapping.cs
@@ -46,10 +46,10 @@ namespace WireMock
         public bool IsStartState => Scenario == null || Scenario != null && NextState != null && ExecutionConditionState == null;
 
         /// <inheritdoc cref="IMapping.IsAdminInterface" />
-        public bool IsAdminInterface => Provider is DynamicResponseProvider || Provider is DynamicAsyncResponseProvider;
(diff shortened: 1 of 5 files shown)


src/WireMock.Net/Owin/WireMockMiddleware.cs
```diff
@@ -99,7 +100,7 @@ private async Task InvokeInternal(IContext ctx)
                     return;
                 }
 
-                logRequest = !targetMapping.IsAdminInterface;
+                logRequest = !targetMapping.IsAdminInterface || targetMapping.Provider is ProxyAsyncResponseProvider;
```
StefH: The mapping class already has this code:
``` c#
public bool IsAdminInterface => Provider is DynamicResponseProvider || Provider is DynamicAsyncResponseProvider || Provider is ProxyAsyncResponseProvider;
```
So no need to add this here?

vitaliydavydiak: That's the root cause of the problem (there is negation). In this check it is actually assumes that proxy response is NOT admin interface in order to log it. I have tried to remove that condition on mapping but it actually conflicts with Admin interface. There is a test for that - FluentMockServer_Proxy_Should_Not_overrule_AdminMappings. So it is kind of workaround to still log proxy requests.

StefH: Ah yes, Sorry.

In that case. Maybe it's better to to add a property to the mapping class, like this:

``` c#
public bool LogRequest => !(Provider is DynamicResponseProvider || Provider is DynamicAsyncResponseProvider).
```

vitaliydavydiak: What about smth like:
`
public bool IsProxyRequest => Provider is ProxyAsyncResponseProvider
`

StefH: But I then you still need to check if it's not an admin request.

vitaliydavydiak: In the MiddleWare - yes. Let me update the PR

src/WireMock.Net/Mapping.cs
```diff
@@ -46,7 +46,10 @@ public class Mapping : IMapping
         public bool IsStartState => Scenario == null || Scenario != null && NextState != null && ExecutionConditionState == null;
 
         /// <inheritdoc cref="IMapping.IsAdminInterface" />
-        public bool IsAdminInterface => Provider is DynamicResponseProvider || Provider is DynamicAsyncResponseProvider || Provider is ProxyAsyncResponseProvider;
+        public bool IsAdminInterface => Provider is DynamicResponseProvider || Provider is DynamicAsyncResponseProvider;
```
StefH: It is too *dangerous* to update this code because `.IsAdminInterface` is used in more places in the application.

I think the best solution is to add a property `LogRequest` as proposed earlier.

vitaliydavydiak: It actually logs both, request and response. Is `LogMapping` better name for this property?

StefH: LogMapping is fine.

StefH:
Thanks.

Issue #350, opened 2019-09-30
Admin requests fail when content type includes a charset


paulroub:
When POSTing a mapping request from C#, by default a "charset=utf-8" parameter is added to the Content-Type header. This causes the request to fail with 

> {"Status":"No matching mapping found"}

Remove the charset and the request succeeds.

Easily reproducible with curl. This succeeds:

```sh
URL=http://localhost:9091/__admin/mappings

JSON='{
    "request": {
        "method": "GET",
        "url": "/some/thing"
    },
    "response": {
        "status": 200,
        "body": "Hello world!",
        "headers": {
            "Content-Type": "text/plain"
        }
    }
}'

curl \
    -H "Content-Type: application/json" \
    -d "$JSON" \
    -iv \
    "$URL"
```

whereas this fails:

```sh
URL=http://localhost:9091/__admin/mappings

JSON='{
    "request": {
        "method": "GET",
        "url": "/some/thing"
    },
    "response": {
        "status": 200,
        "body": "Hello world!",
        "headers": {
            "Content-Type": "text/plain"
        }
    }
}'

curl \
    -H "Content-Type: application/json; charset=utf-8" \
    -d "$JSON" \
    -iv \
    "$URL"
```

StefH:
You can try MyGet version:
WireMock.Net.1.0.32-ci-11942.nupkg

StefH:
@paulroub did you have time to test?

paulroub:
Yes, sorry - all good!

PR #351, opened 2019-09-30
In admin API, match content type JSON with or without charset

paulroub:
Tests, tentative fix for #350.


StefH:
Thanks.

I was trying to get the regex correct.

I will probably reuse your idea.

And could this new contenttypematcher also directly inherited the regexmatcher? Just like the wildcardmatcher?

paulroub:
I don’t see why not. I don’t really know the codebase, so I’m sure there
are better ways to implement my idea.

PR #358, opened 2019-10-08
Fix JsonMatcher (parsing DateTimeOffset)

StefH:


src/WireMock.Net/Transformers/HandleBarsJsonPath.cs
```diff
@@ -1,14 +1,20 @@
...
 using WireMock.Validation;
 
 namespace WireMock.Transformers
 {
     internal static class HandleBarsJsonPath
     {
+        private static readonly JsonSerializerSettings JsonSerializerSettings = new JsonSerializerSettings()
+        {
+            DateParseHandling = DateParseHandling.None
+        };
```
JoshKeegan: All looks good. 

Only thing to consider is whether this could/should be common to both by moving it into `JsonUtils.Parse`?
I can't think of why `JsonMatch` would need to deserialise one way and `JsonPath` another, but there could be a good reason they're done separately? :)

StefH: Yes. Better to use the JsonUtils.Parse to always use DateParseHandling.None.

PR #366, opened 2019-10-22
Update ObsoleteAnnotations

StefH:


src/WireMock.Net/Server/FluentMockServer.cs
```diff
@@ -25,7 +25,7 @@ namespace WireMock.Server
     /// <summary>
     /// The fluent mock server.
     /// </summary>
-    [Obsolete("Will be replaced by WireMockServer in version 1.1.0")]
+    [Obsolete("Will be replaced by WireMockServer in version 2.x.x")]
```
prestonj: @StefH Just a note that adding the `Obsolete` annotation here will cause projects using the current version of WireMock (and of course the obsoleted class) to fail to compile when warning levels are turned up and `treat warnings as errors` is turned on. Since the `WireMockServer` class does not yet exist (at least in version 1.0.33), and our team is reluctant to change our project settings, there was no way to get around this but to downgrade our version of WireMock to pre obsolete annotation.

PR #398, opened 2020-01-08
Feature/xpath transformer

### LINKED ISSUE #397: Question/Feature: Add support for selecting XPath in response template
Hi,
I am looking to transform soap response. And it would be good if i can use the XPath to select values from request. I believe it can be done using regex but XPath would be a natural fit for this use case.

Question: Is it already supported?

If not happy to do a PR. It would be similar to how json template work.
"{{XPath.Select request.body \"/todo-list/todo-item[1]/text()\"}}") to return text of 1st element in response.

kashifsoofi:
Fixes #397 Added feature to select values from xml using XPath.
Implemented following as suggested

- `XPath.SelectSingleNode`
- `XPath.SelectNodes`
Also implemented
- `XPath.Evaluate`

Reason for using OuterXml in `HandleBarsXPath.cs` is as follows but happy to change :
Testing
```
<todo-list>
  <todo-item id='a1'>abc</todo-item>
  <todo-item id='a2'>def</todo-item>
  <todo-item id='a3'>xyz</todo-item>
</todo-list>
```
with `/todo-list/todo-item[1]` results in `<todo-item id="a1">abc</todo-item>`
 using https://www.freeformatter.com/xpath-tester.html. I think it is the correct behaviour as to select text I would provide `/todo-list/todo-item[1]/text()`. However same experssion return only the text using XPath2SelectSingleNode.

### DIFF
--- a/src/WireMock.Net/Transformers/HandleBarsHelpers.cs
+++ b/src/WireMock.Net/Transformers/HandleBarsHelpers.cs
@@ -17,6 +17,8 @@ namespace WireMock.Transformers
 
             HandleBarsXeger.Register(handlebarsContext);
 
+            HandleBarsXPath.Register(handlebarsContext);
+
             HandleBarsFile.Register(handlebarsContext, fileSystemHandler);
         }
     }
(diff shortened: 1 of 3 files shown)


StefH:
Looks good.

Did you also test this new code in a real application/integration test ?

kashifsoofi:
> Looks good.
> 
> Did you also test this new code in a real application/integration test ?

Not yet.

kashifsoofi:
> 
> 
> > Looks good.
> > Did you also test this new code in a real application/integration test ?
> 
> Not yet.

Tested and its working as expected.

PR #409, opened 2020-01-30
Admin Delete with mappings in body

thewholuver94:
Using HTTP `Delete `verb, a user can send multiple mapping jsons (with GUIDs) for deletion.
Note: `AllowBodyForAllHttpMethods `must be set to true.

### DIFF
--- a/src/WireMock.Net/Server/WireMockServer.Admin.cs
+++ b/src/WireMock.Net/Server/WireMockServer.Admin.cs
@@ -555,7 +555,6 @@ namespace WireMock.Server
 
         private ResponseMessage MappingsDelete(RequestMessage requestMessage)
         {
-            // if we allow body for HTTP Delete & a body is defined
             if (!string.IsNullOrEmpty(requestMessage.Body))
             {
                 var deletedGuids = MappingsDeleteMappingFromBody(requestMessage);
--- a/src/WireMock.Net/Server/WireMockServer.Admin.cs
+++ b/src/WireMock.Net/Server/WireMockServer.Admin.cs
@@ -566,7 +566,7 @@ namespace WireMock.Server
                 else
                 {
                     // return bad request
-                    return ResponseMessageBuilder.Create($"Poorly formed mapping JSON.", 400);
+                    return ResponseMessageBuilder.Create("Poorly formed mapping JSON.", 400);
                 }
             }
             else
--- a/src/WireMock.Net/Server/WireMockServer.Admin.cs
+++ b/src/WireMock.Net/Server/WireMockServer.Admin.cs
@@ -558,9 +558,8 @@ namespace WireMock.Server
             // if we allow body for HTTP Delete & a body is defined
             if (_settings.AllowBodyForAllHttpMethods.HasValue && _settings.AllowBodyForAllHttpMethods.Value && !String.IsNullOrEmpty(requestMessage.Body))
             {
-                List<Guid> deletedGuids = new List<Guid>();
-                bool success = MappingsDeleteMappingFromBody(requestMessage, ref deletedGuids);
-                if (success)
+                IEnumerable<Guid> deletedGuids = MappingsDeleteMappingFromBody(requestMessage);
+                if (deletedGuids != null)
                 {
                     return ResponseMessageBuilder.Create($"Mappings deleted. Affected GUIDs: [{String.Join(", ", deletedGuids?.ToArray())}]");
                 }
@@ -580,8 +579,10 @@ namespace WireMock.Server
             }
         }
 
-        private bool MappingsDeleteMappingFromBody(RequestMessage requestMessage, ref List<Guid> deletedGuids)
+        private IEnumerable<Guid> MappingsDeleteMappingFromBody(RequestMessage requestMessage)
         {
+            List<Guid> deletedGuids = new List<Guid>();
+
             try
             {
                 var mappingModels = DeserializeRequestMessageToArray<MappingModel>(requestMessage);
@@ -603,15 +604,15 @@ namespace WireMock.Server
             catch (ArgumentException a)
             {
                 _settings.Logger.Error("ArgumentException: {0}", a);
-                return false;
+                return null;
             }
             catch (Exception e)
             {
                 _settings.Logger.Error("Exception: {0}", e);
-                return false;
+                return null;
             }
 
-            return true;
+            return deletedGuids;
         }
 
         private ResponseMessage MappingsReset(RequestMessage requestMessage)
--- a/src/WireMock.Net/Server/WireMockServer.Admin.cs
+++ b/src/WireMock.Net/Server/WireMockServer.Admin.cs
@@ -555,11 +555,63 @@ namespace WireMock.Server
 
         private ResponseMessage MappingsDelete(RequestMessage requestMessage)
         {
-            ResetMappings();
+            // if we allow body for HTTP Delete & a body is defined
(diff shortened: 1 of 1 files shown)


thewholuver94:
Hi again @StefH 

I've extended functionality of the HTTP Delete verb at the` __admin/mappings` endpoint by allowing the user to send a json body with his mappings for deletion. The mappings must include Guids. Each Guid which is mapped to a valid mapping on the server is deleted and confirmed in the response.

If the mappings json is malformed, the user receives a BadRequest response.

Note that the `AllowBodyForAllHttpMethods` must be set to true.

src/WireMock.Net/Server/WireMockServer.Admin.cs
```diff
@@ -555,11 +555,63 @@ private ResponseMessage MappingsPost(RequestMessage requestMessage)
...
+                ResetMappings();
 
-            ResetScenarios();
+                ResetScenarios();
+
+                return ResponseMessageBuilder.Create("Mappings deleted");
+            }
+        }
+
+        private bool MappingsDeleteMappingFromBody(RequestMessage requestMessage, ref List<Guid> deletedGuids)
```
StefH: Return `ICollection<Guid>` or  `IEnumerable<Guid>` in case it's ok.
Else return `null` when something throws.

src/WireMock.Net/Server/WireMockServer.Admin.cs
```diff
@@ -555,11 +555,63 @@ private ResponseMessage MappingsPost(RequestMessage requestMessage)
 
         private ResponseMessage MappingsDelete(RequestMessage requestMessage)
         {
-            ResetMappings();
+            // if we allow body for HTTP Delete & a body is defined
+            if (_settings.AllowBodyForAllHttpMethods.HasValue && _settings.AllowBodyForAllHttpMethods.Value && !String.IsNullOrEmpty(requestMessage.Body))
```
StefH: For admin commands, it's not needed to check for `AllowBodyForAllHttpMethods`. It's ok to use this functionality.

src/WireMock.Net/Server/WireMockServer.Admin.cs
```diff
@@ -555,11 +555,63 @@ private ResponseMessage MappingsPost(RequestMessage requestMessage)
...
+                List<Guid> deletedGuids = new List<Guid>();
+                bool success = MappingsDeleteMappingFromBody(requestMessage, ref deletedGuids);
+                if (success)
+                {
+                    return ResponseMessageBuilder.Create($"Mappings deleted. Affected GUIDs: [{String.Join(", ", deletedGuids?.ToArray())}]");
+                }
+                else
+                {
+                    // return bad request
+                    return ResponseMessageBuilder.Create($"Poorly formed mapping JSON.", 400);
```
StefH: no need for $

src/WireMock.Net/Server/WireMockServer.Admin.cs
```diff
@@ -555,11 +555,64 @@ private ResponseMessage MappingsPost(RequestMessage requestMessage)
 
         private ResponseMessage MappingsDelete(RequestMessage requestMessage)
         {
-            ResetMappings();
+            // if we allow body for HTTP Delete & a body is defined
+            if (!String.IsNullOrEmpty(requestMessage.Body))
```
StefH: Just use `string` everywhere in file.

src/WireMock.Net/Server/WireMockServer.Admin.cs
```diff
@@ -555,11 +555,64 @@ private ResponseMessage MappingsPost(RequestMessage requestMessage)
...
+            {
+                IEnumerable<Guid> deletedGuids = MappingsDeleteMappingFromBody(requestMessage);
```
StefH: just use `var deletedGuids = `

src/WireMock.Net/Server/WireMockServer.Admin.cs
```diff
@@ -555,11 +555,64 @@ private ResponseMessage MappingsPost(RequestMessage requestMessage)
...
-            ResetScenarios();
+                ResetScenarios();
+
+                return ResponseMessageBuilder.Create("Mappings deleted");
+            }
+        }
+
+        private IEnumerable<Guid> MappingsDeleteMappingFromBody(RequestMessage requestMessage)
+        {
+            List<Guid> deletedGuids = new List<Guid>();
```
StefH: Just use `var deletedGuids = new List<Guid>();`

src/WireMock.Net/Server/WireMockServer.Admin.cs
```diff
@@ -555,11 +555,64 @@ private ResponseMessage MappingsPost(RequestMessage requestMessage)
...
         private ResponseMessage MappingsDelete(RequestMessage requestMessage)
         {
-            ResetMappings();
+            // if we allow body for HTTP Delete & a body is defined
+            if (!String.IsNullOrEmpty(requestMessage.Body))
+            {
+                IEnumerable<Guid> deletedGuids = MappingsDeleteMappingFromBody(requestMessage);
+                if (deletedGuids != null)
+                {
+                    return ResponseMessageBuilder.Create($"Mappings deleted. Affected GUIDs: [{String.Join(", ", deletedGuids?.ToArray())}]");
```
StefH: `?.` is not needed because you already do a null check two lines above.

src/WireMock.Net/Server/WireMockServer.Admin.cs
thewholuver94: Just curious, what are the advantages of declaring ```deletedGuids``` var as opposed to IEnumerable<Guid> ?

StefH: I normally use always var for complex objects and lists/enumerable and so on. (Except when the logic is difficult to understand.

Only for primitive types like bool, int, string.

src/WireMock.Net/Server/WireMockServer.Admin.cs
```diff
@@ -555,11 +555,64 @@ private ResponseMessage MappingsPost(RequestMessage requestMessage)
 
         private ResponseMessage MappingsDelete(RequestMessage requestMessage)
         {
-            ResetMappings();
+            // if we allow body for HTTP Delete & a body is defined
```
StefH: This comment can be removed.

thewholuver94:
@StefH - How can I update the Wiki to reflect the changes?

StefH:
I did send an invite to this project, you should be able to edit the wiki I think.

StefH:
I just remebered..

This file https://github.com/WireMock-Net/WireMock.Net/blob/master/src/WireMock.Net.RestClient/IWireMockAdminApi.cs
needs also an update to support this new method.

thewholuver94:
I think I just made a mess of the repo while trying to fix this. I don't want to mess it up anymore, can you take a look?

thewholuver94:
I committed 2dcf4b3104623109aa6a9acb1014bdffcae987e5 directly to the upstream by accident.

StefH:
Sorry, I forgot to tell you. I did update some code in your branch.

I'll take a look a try to fix.

thewholuver94:
> I'll take a look a try to fix.

Thank for having a look.

>Sorry, I forgot to tell you. I did update some code in your "PathUtils" branch.

I saw, that's totally fine. Let's move this part of the discussion to that PR.

StefH:
@thewholuver94 --> I did remove all your commits from the `master` branch. It should be good to go.

PR #413, opened 2020-02-02
Fix new Delete with body missing from IWireMockAdminApi interface

thewholuver94:
As requested.


thewholuver94:
I didn't realize that was reverted too. I resubmitted the code.

thewholuver94:
@StefH can you please review the PR? Thanks

thewholuver94:
> This change does not work correctly, because the body is not parsed in this line:
> https://github.com/WireMock-Net/WireMock.Net/blob/3250604b5a56358838217ff7ef17cde7c57d429a/src/WireMock.Net/Owin/Mappers/OwinRequestMapper.cs#L50
> 
> (DELETE has no body semantics, so this line will not parse body)
> So maybe skip this PR or change it so that your just provide all GUID's in the query string?

I'm not sure what you mean by not working correctly? I tested it and it works fine... as long as `AllowBodyForAllHttpMethods` is set to true then the body arrives just fine along with the request to `private ResponseMessage MappingsDelete(RequestMessage requestMessage)`.

Also, before opening this PR I looked into whether DELETE can have body semantics and found [this documentation saying it may have a body.](https://developer.mozilla.org/en-US/docs/Web/HTTP/Methods/DELETE)

Am I missing something?

thewholuver94:
And also [in the RFC, it is not forbidden.](https://tools.ietf.org/html/rfc7231#section-4.3.5) It only says that `sending a payload body on a DELETE request might cause some existing implementations to reject the request.`. But we know this not to be the case in the WireMock.Net implementation...

StefH:
Currently, the logic form WireMock is that DELETE cannot have a body:
https://github.com/WireMock-Net/WireMock.Net/blob/3250604b5a56358838217ff7ef17cde7c57d429a/src/WireMock.Net/Util/BodyParser.cs#L19

Because of this logic, all requests (normal + admin) will not allow a body for delete. This behaviour can indeed be changed by AllowBodyForAllHttpMethods, however this settings is actually meant for normal requests, for admin requests, this should actually not be applied (for admin requests: all is allowed)
This difference in logic is not yet implemented, the OwinRequestMapper.cs maps the request, without knowing if it's a admin request.

So a fix could be:
1. allow body for all delete
2. add special logic to the OwinRequestMapper.cs to determine if it's an admin-request or normal request. Or move the ShouldParseBody check to a later stage. (Maybe MappingMatcher..?)

thewholuver94:
> 1. allow body for all delete

Then I think that this should be the fix since the RFC indicates that body is allowed for DELETE.
It also happens to be the simplest fix.
Do you agree?

StefH:
You also need to update some other unit-tests because of this update.

thewholuver94:
Yes I see that now. Thanks I'll get to it soon.

thewholuver94:
Stef can you please rerun the CI? Not sure why it didn't pass.

test/WireMock.Net.Tests/WireMockServer.Admin.cs
```diff
@@ -404,5 +405,53 @@ public void WireMockServer_Admin_AddMappingsAndSaveToFile()
...
             staticMappingHandlerMock.Verify(m => m.WriteMappingFile(Path.Combine("folder", guid + ".json"), It.IsAny<string>()), Times.Once);
         }
+
+        [Fact]
+        public async void WireMockServer_Admin_DeleteMappings()
+        {
+            // Arrange
+            var server = WireMockServer.Start(new WireMockServerSettings
+            {
+                AllowBodyForAllHttpMethods = true,
```
StefH: This can be removed I think?

test/WireMock.Net.Tests/WireMockServer.Admin.cs
```diff
@@ -404,5 +405,53 @@ public void WireMockServer_Admin_AddMappingsAndSaveToFile()
...
+                StartAdminInterface = true,
+                ReadStaticMappings = false,
+                WatchStaticMappings = false,
+                WatchStaticMappingsInSubdirectories = false
+            });
+            string folder = Path.Combine(GetCurrentFolder(), "__admin", "mappings", "delete_mappings");
+            server.ReadStaticMappings(folder);
```
StefH: Instead of reading static mapping in this test, I think it's better to just add 3 new mappings to the server and then delete 2 mappings using tghe new delete method..

test/WireMock.Net.Tests/WireMockServer.Admin.cs
```diff
@@ -404,5 +405,53 @@ public void WireMockServer_Admin_AddMappingsAndSaveToFile()
...
+            var response = await new HttpClient().SendAsync(request);
+
+            // Assert
+            Check.That(server.MappingModels).HasSize(0);
+            Check.That(response.StatusCode).Equals(HttpStatusCode.OK);
+            Check.That(await response.Content.ReadAsStringAsync()).Equals("{\"Status\":\"Mappings deleted. Affected GUIDs: [00000002-ee28-4f29-ae63-1ac9b0802d86, 00000002-ee28-4f29-ae63-1ac9b0802d87]\"}");
+        }
+
+        private string GetRequestBodyFromJsonsInFolder(string folder)
+        {
```
StefH: This is helper is not needed when just adding 3 new mappings in code.

test/WireMock.Net.Tests/WireMock.Net.Tests.csproj
```diff
@@ -77,6 +77,12 @@
     <None Update="__admin\mappings\array.json">
       <CopyToOutputDirectory>PreserveNewest</CopyToOutputDirectory>
     </None>
+    <None Update="__admin\mappings\delete_mappings\00000002-ee28-4f29-ae63-1ac9b0802d86.json">
+      <CopyToOutputDirectory>PreserveNewest</CopyToOutputDirectory>
+    </None>
```
StefH: Not needed, see comments below in other files.

test/WireMock.Net.Tests/WireMockServerTests.cs
```diff
@@ -160,9 +160,8 @@ public async Task WireMockServer_Should_exclude_restrictedResponseHeader()
             Check.That(response.Headers.Contains("Transfer-Encoding")).IsFalse();
         }
 
-        [Theory]
-        [InlineData("DELETE")]
 #if !NET452
+        [Theory]
```
StefH: I think this is not ok. Now the [Theory] is not defined for .net 452? This should result into a build-error? (BTW I cannot remember anymore why this is?)

Best to replace code like:
``` c#
[Theory]
[InlineData("PUT")]
[InlineData("OPTIONS")]
[InlineData("REPORT")]
[InlineData("DELETE")]
#if !NET452
[InlineData("TRACE")]
[InlineData("GET")]
#endif
```

thewholuver94: The issue is that I can't leave the [Theory] tag outside the if region since _that_ results in a build error because I can't have the [Theory] tag without an InlineData tag.

I'm building against 461 and I don't get any build error.

thewholuver94: @StefH I think this unit test is problematic at best. It looks like nothing inside the #if region is tested.
And once I took the tags out of the region & compiled using 461 (NOT 452), the unit test still failed.

I got an exception on `GET` as follows: `System.Net.ProtocolViolationException : Cannot send a content-body with this verb-type.`

For `TRACE`, the unit test failed - it returned a `BadRequest` instead of expected `OK`.

Maybe this unit test needs to be redesigned to not use `HttpRequestMessage` since we need to test a situation when a body actually arrives to the server.

Another possible solution would to allow body on all requests by using `System.Net.Http.WinHttpHandler` as [indicated here.](https://stackoverflow.com/a/47902348/10233008)

StefH: What happens if you put exactly the code like:
``` c#
[Theory]
[InlineData("PUT")]
[InlineData("OPTIONS")]
[InlineData("REPORT")]
[InlineData("DELETE")]
#if !NET452
[InlineData("TRACE")]
[InlineData("GET")]
#endif
```

It should run all verbs for net-core and net46. And skip those 2 problems verbs for net452

thewholuver94: It fails, and ignores the 2 problem verbs even when using net461.

The failure here is because the 4 other verbs do allow the body, and then are caught by the first mapping which returns 400 BadRequest. This unit test specifically tests methods where body is not allowed, so including the 4 verbs where body is allowed will obviously fail the test.

StefH: Sorry. Now I understand, there are 2 test methods...

OK.

**Fix 1:**
`WireMockServer_Should_exclude_body_for_methods_where_body_is_definitely_disallowed` should be completely surrounded by if-def NET452 in your code.

like:
``` c#
#if !NET452
        [Theory]
        [InlineData("TRACE")]
        [InlineData("GET")]
        public async Task WireMockServer_Should_exclude_body_for_methods_where_body_is_definitely_disallowed(string method)
        {
            // Assign
            ...
        }
#endif
```

**Fix 2:**
For test `WireMockServer_Should_not_exclude_body_for_supported_methods` : the "DELETE" should be added ( I think you already did this)

thewholuver94: Maybe I'm not building the project the correct way...
When these are my settings, everything inside the #if region does not show up in the test explorer... 

image

StefH: Maybe I did remove the net452 Target from the test project.

So if you fix my last two things, the PR is ready.

thewholuver94: Just pushed Fix 1. 
Fix 2 I already did as you mentioned.

Resolving this thread.

Issue #418, opened 2020-02-13
Body matching fails if body has newline


thewholuver94:
When using `WildcardMatcher`, even if using the following config in json mapping file:
`"Body": {
      "Matcher": {
        "Name": "WildcardMatcher",
        "Pattern": "*"
      }
`
the mapping is never triggered if I have a newline in the request body. After removing the newline, the mapping is always triggered.

PR #419, opened 2020-02-13
Support multi line wild card matching

thewholuver94:
simple fix - just allow multi line regex matching.


thewholuver94:
Fixes #418

StefH:
Thanks

PR #434, opened 2020-03-03
Option to disable JSON deserialization

sebastianmattar:
I had problems mocking a system with very specific requirements concerning JSON formatting. To make this work, I implemented an option to disable JSON reformatting.

### DIFF
--- a/src/WireMock.Net/Util/BodyParser.cs
+++ b/src/WireMock.Net/Util/BodyParser.cs
@@ -107,7 +107,7 @@ namespace WireMock.Util
             return BodyType.Bytes;
         }
 
-        public static async Task<BodyData> Parse([NotNull] Stream stream, [CanBeNull] string contentType, bool deserializeJson)
+        public static async Task<BodyData> Parse([NotNull] Stream stream, [CanBeNull] string contentType, bool deserializeJson = true)
         {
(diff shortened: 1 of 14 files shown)


src/WireMock.Net/Util/BodyParser.cs
```diff
@@ -107,7 +107,7 @@ public static BodyType DetectBodyTypeFromContentType([CanBeNull] string contentT
             return BodyType.Bytes;
         }
 
-        public static async Task<BodyData> Parse([NotNull] Stream stream, [CanBeNull] string contentType)
+        public static async Task<BodyData> Parse([NotNull] Stream stream, [CanBeNull] string contentType, bool deserializeJson)
```
StefH: Maybe better to change the order from this parameter because contentType is optional?
Like
`Parse([NotNull] Stream stream, bool deserializeJson, [CanBeNull] string contentType)`

or even better , set it default to true:
`Parse([NotNull] Stream stream, bool deserializeJson = true, [CanBeNull] string contentType = null)`

sebastianmattar: I changed it to 
`Parse([NotNull] Stream stream, [CanBeNull] string contentType = null, bool deserializeJson = true)`
Hope this is ok?

PR #438, opened 2020-03-12
Change the nuget license string to Apache-2.0 to match LICENSE file.

ml-workshare:
Hi guys. Great library. Thank you for the excellent work!

We noticed an inconsistency with the licensing terms on the github repo - nuget.org reports MIT, but the code itself is Apache. Could you provide some clarification as to which is intended? 

Thanks!

### DIFF
--- a/src/WireMock.Net.Abstractions/WireMock.Net.Abstractions.csproj
+++ b/src/WireMock.Net.Abstractions/WireMock.Net.Abstractions.csproj
@@ -12,7 +12,7 @@
     <PackageReleaseNotes>See CHANGELOG.md</PackageReleaseNotes>
     <PackageIconUrl>https://raw.githubusercontent.com/WireMock-Net/WireMock.Net/master/WireMock.Net-Logo.png</PackageIconUrl>
     <PackageProjectUrl>https://github.com/WireMock-Net/WireMock.Net</PackageProjectUrl>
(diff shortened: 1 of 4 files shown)


StefH:
Good point.

I would like to keep this project MIT, because that's the most `relaxed` license.
However, I copied some code from mock4net, which has Apache2. So maybe I'm not allowed to switch to MIT?

I'm not an expert on this.

Can you provide some help?

ml-workshare:
I'm not really sure to be honest. I wouldn't think so. I doubt you'd be able to overwrite an existing license with a more permissive license, and certainly all the copyright information should remain. https://www.quora.com/Does-the-apache-license-allow-me-to-modify-the-program

StefH:
BTW I found this site:
http://vschart.com/compare/mit-license/vs/apache-license

However this not help me very much.

But I think I need to change this library to Apache-2.0 and even include the original license from mock4net in this project.

StefH:
see also https://github.com/WireMock-Net/WireMock.Net/pull/444

StefH:
@ml-workshare Can you take a look at #444 ?

StefH:
@ml-workshare Thank you very much for the help and research.

I'm going to close your PR, because I've added the license to the global file Directory.Build.props file in PR #444

PR #446, opened 2020-03-25
When port is provided: WireMockServer still takes a random port

StefH:


StefH:
https://github.com/WireMock-Net/WireMock.Net/issues/445

PR #473, opened 2020-05-23
Fixed Proxy when using MultipartForm with byte[]

StefH:


src/WireMock.Net/Http/ByteArrayContentHelper.cs
```diff
@@ -0,0 +1,30 @@
...
+{
+    internal static class ByteArrayContentHelper
+    {
+        /// <summary>
+        /// Creates a ByteArrayContent object.
+        /// </summary>
+        /// <param name="content">The byte[] content (cannot be null)</param>
+        /// <param name="contentType">The ContentType (can be null)</param>
+        /// <returns>ByteArrayContent</returns>
+        internal static ByteArrayContent Create([NotNull] byte[] content, [CanBeNull] MediaTypeHeaderValue contentType)
```
LevYas, line 16: I use nullable reference types feature in the modern C#, where it is available, it rocks. Just a note.

StefH: Yes. I started using the new features for some small new projects on GitHub.
However, I did not yet check this project. Will take some time to fix all warnings I think.

LevYas: Yes, it will. And sometimes it reveals some problems or requires a chain of modifications, but, ultimately, you have a good feeling that the codebase became more reliable and organized, I like it :)

PR #483, opened 2020-07-04
Bring in the WireMock.Net.FluentAssertions tests

akamud:
Porting over the tests from the original project.

### DIFF
--- a/test/WireMock.Net.Tests/FluentAssertions/WireMockAssertionsTests.cs
+++ b/test/WireMock.Net.Tests/FluentAssertions/WireMockAssertionsTests.cs
@@ -1,0 +1,62 @@
+﻿using FluentAssertions;
+using System;
+using System.Net.Http;
+using WireMock.RequestBuilders;
+using WireMock.ResponseBuilders;
+using WireMock.Server;
+using Xunit;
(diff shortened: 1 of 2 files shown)


StefH:
Maybe put the tests into separate test files? Based on the name from the class?

akamud:
Can you elaborate more, please? I'm only testing the `WireMockAssertions` at the moment, so it is in the `WireMockAssertionsTests`, but since it depends on the other classes, they are covered as part of the integrated tests I wrote.

Do you mean writing separate test for the supporting classes? (WireMockReceivedAssertions, WireMockANumberOfCallsAssertions, etc.?)

StefH:
Correct.
Unless it does not make sense?

akamud:
I think they are very simple at the moment and writing extra unit tests for them doesn't bring much value, since their behavior are guaranteed by the Assertions tests and can't be mocked at the moment.
The only part I'm not testing yet, is the `Received(int numberOfCalls)`, but that's because I'm not using it either. They were more like a draft for the future, since I wasn't planning on writing the scenarios for asserting exactly the number of times for now.

Given that, would you like me to change anything?

StefH:
I will check tomorrow.

PR #539, opened 2020-11-16
Support for partial JSON matching

gleb-osokin:
Current JsonMatcher only support exact match (via JToken.DeepEquals).
This PR introduces new JsonPartialMatcher, which tolerates input properties, not existing on the Matcher value.

Example:
Matcher value **{"test": "abc"}** against input **{"test":"abc"}** is matched by both JsonMatcher and JsonPartialMatcher
Matcher value **{"test":"abc"}** against input **{"test":"abc","other":"xyz"}** is not matched by JsonMatcher, but is matched by JsonPartialMatcher

One side benefit of this PR is that JsonMatcher (and JsonPartialMatcher) are now extensible via virtual AreEqual(JToken, JToken) method.


StefH (changes requested):
Thanks,

You also need to add this new matcher to 
https://github.com/WireMock-Net/WireMock.Net/blob/c02dbeb5eea5fd01ce657e8ec362fd2dddcc5b19/src/WireMock.Net/Serialization/MatcherMapper.cs

And if possible also add an unit-test for that.

gleb-osokin:
Thanks @StefH !

Added changes as suggested.

src/WireMock.Net/Matchers/JsonPartialMatcher.cs
```diff
@@ -0,0 +1,86 @@
...
+            return IsMatch(value, input);
+        }
+
+        private static bool IsMatch(JToken value, JToken token)
+        {
+            if (value == null || value == token)
+                return true;
+
+            if (token == null)
+                return false;
```
StefH, line 62: Can you please use curly brackets for if-statements?

src/WireMock.Net/Matchers/JsonPartialMatcher.cs
```diff
@@ -0,0 +1,86 @@
...
+                case JTokenType.Object:
+                    var nestedValues = value.ToObject<Dictionary<string, JToken>>();
+                    return nestedValues?.Any() != true ||
+                           nestedValues.All(pair => IsMatch(pair.Value, token.SelectToken(pair.Key)));
+
+                case JTokenType.Array:
+                    var filtersArray = value.ToObject<JToken[]>();
+                    var tokenArray = token.ToObject<JToken[]>();
+
+                    if (filtersArray?.Any() != true)
```
StefH, line 75: Can you please use curly brackets for if-statements?

StefH (changes requested):
some more small things, can you please take a look?

src/WireMock.Net/Matchers/JsonPartialMatcher.cs
gleb-osokin: Sure, added.

Thanks for reviewing this!

test/WireMock.Net.Tests/Matchers/JsonPartialMatcherTests.cs
```diff
@@ -0,0 +1,356 @@
...
+            // Act
+            double match = matcher.IsMatch(input);
+
+            // Assert
+            Assert.Equal(0.0, match);
+        }
+
+        public static IEnumerable<object[]> ValidMatches()
+        {
+            yield return new object[] { "{}", "\"test\"" };
```
StefH, line 330: I'm a bit confused, why would this be a 1.0 match?

gleb-osokin: Good catch, thanks. Fixed and moved it to Invalid cases.

StefH (commented):
Can you please exlain?

test/WireMock.Net.Tests/Matchers/JsonPartialMatcherTests.cs
```diff
@@ -0,0 +1,356 @@
...
+            yield return new object[] { "\"test\"", "\"test\"" };
+            yield return new object[] { "123", "123" };
+            yield return new object[] { "[\"test\"]", "[\"test\"]" };
+            yield return new object[] { "[\"test\"]", "[\"test\", \"test1\"]" };
+            yield return new object[] { "[123]", "[123]" };
+            yield return new object[] { "{ \"test\":\"value\" }", "{\"test\":\"value\"}" };
+            yield return new object[] { "{ \"test.test1\":\"value\" }", "{\"test\":{\"test1\":\"value\"}}" };
+            yield return new object[] { "{\"test\":{\"test1\":\"value\"}}", "{\"test\":{\"test1\":\"value\"}}" };
+            yield return new object[] { "[{ \"test.test1\":\"value\" }]", "[{\"test\":{\"test1\":\"value\"}}]" };
```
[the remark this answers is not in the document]
gleb-osokin: My idea here was to allow shorthand matching against nested objects.
E.g.:
**{ "test.nested": "abc" }** matches against **{ "test": { "nested": "abc" } }**

Which means the property here is parsed as a JPath.
I've added a few cases, which treat properties containing dots.

test/WireMock.Net.Tests/Matchers/JsonPartialMatcherTests.cs
```diff
@@ -0,0 +1,356 @@
...
+            var matcher = new JsonPartialMatcher(value);
+
+            // Act
+            double match = matcher.IsMatch(input);
+
+            // Assert
+            Assert.Equal(0.0, match);
+        }
+
+        public static IEnumerable<object[]> ValidMatches()
```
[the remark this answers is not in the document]
gleb-osokin: Sure, added. I've also changed some inputs to make it more clear.

test/WireMock.Net.Tests/Matchers/JsonPartialMatcherTests.cs
```diff
@@ -0,0 +1,360 @@
...
+        {
+            yield return new object[] { "{\"test\":\"abc\"}", "{\"test\":\"abc\",\"other\":\"xyz\"}" };
+            yield return new object[] { "\"test\"", "\"test\"" };
+            yield return new object[] { "123", "123" };
+            yield return new object[] { "[\"test\"]", "[\"test\"]" };
+            yield return new object[] { "[\"test\"]", "[\"test\", \"other\"]" };
+            yield return new object[] { "[123]", "[123]" };
```
StefH, line 335: Can you also add example like:
`yield return new object[] { "[123]", "[123, 456]" };`

test/WireMock.Net.Tests/Matchers/JsonPartialMatcherTests.cs
```diff
@@ -0,0 +1,360 @@
...
+            yield return new object[] { "{ \"test\":\"value\" }", "{\"test\":\"value\",\"other\":123}" };
+            yield return new object[] { "{ \"test\":\"value\" }", "{\"test\":\"value\"}" };
+            yield return new object[] { "{ \"test.nested\":\"value\" }", "{\"test\":{\"nested\":\"value\"}}" };
```
StefH, line 338: How is it possible that a dot-notation equals the nested value?

StefH (changes requested):

test/WireMock.Net.Tests/Matchers/JsonPartialMatcherTests.cs
gleb-osokin: Sure, added!

test/WireMock.Net.Tests/Matchers/JsonPartialMatcherTests.cs
gleb-osokin: Again, the idea here is to allow shorthand matching against nested objects.
E.g.: **{ "test.nested": "abc" }** matches against { "test": { "nested": "abc" } }

The property here, essentially, is treated as a JPath, like in the JSONPathParser.

I've renamed the JsonMatcher/JsonPartialMatcher.AreEqual protected method into IsMatch to avoid the confusion.

StefH: Can you please create a separate Unit test which only tests these shorthand notations. (because I guess I will forget this logic in a year from now why this works...)

gleb-osokin: Good point. Moved JPath tests into distinct tests. Hope it's better now.

StefH (approved):
Thanks

PR #557, opened 2020-12-16
Implement random delays

myarichuk:
In order to use WireMock as a proxy for chaos testing, introducing random delays would be a useful feature. This is my attempt implementing random delays.

Note: I tried to make as less changes as possible to the code, as such change can be introduced in a couple of ways, I think.

### DIFF
--- a/src/WireMock.Net/ResponseBuilders/Response.cs
+++ b/src/WireMock.Net/ResponseBuilders/Response.cs
@@ -5,6 +5,7 @@ using System.Collections.Generic;
 using System.Linq;
 using System.Net;
 using System.Text;
+using System.Threading;
 using System.Threading.Tasks;
 using JetBrains.Annotations;
 using WireMock.Proxy;
@@ -22,10 +23,32 @@ namespace WireMock.ResponseBuilders
     /// </summary>
     public partial class Response : IResponseBuilder
     {
+
+        private static readonly ThreadLocal<Random> Random = new ThreadLocal<Random>(() => new Random(DateTime.UtcNow.Millisecond));
+
+        private int _minDelayMilliseconds;
+        private int _maxDelayMilliseconds;
+        private bool _hasRandomDelay;
+
+        private TimeSpan? _delay;
+
         /// <summary>
         /// The delay
         /// </summary>
-        public TimeSpan? Delay { get; private set; }
+        public TimeSpan? Delay
+        {
+            get
+            {
+                if (_hasRandomDelay)
+                    return TimeSpan.FromMilliseconds(Random.Value.Next(_minDelayMilliseconds, _maxDelayMilliseconds));
+
+                return _delay;
+            }
+            private set
+            {
+                _delay = value;
+            }
+        }
 
         /// <summary>
         /// Gets a value indicating whether [use transformer].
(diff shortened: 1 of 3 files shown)


src/WireMock.Net/ResponseBuilders/Response.cs
```diff
@@ -22,10 +23,31 @@ namespace WireMock.ResponseBuilders
...
+
         /// <summary>
         /// The delay
         /// </summary>
-        public TimeSpan? Delay { get; private set; }
+        public TimeSpan? Delay
+        {
+            get
+            {
+                if (_hasRandomDelay)
```
StefH, line 41: Use `{` and `}` here.

src/WireMock.Net/ResponseBuilders/Response.cs
```diff
@@ -22,10 +23,31 @@ namespace WireMock.ResponseBuilders
...
+
+        private int _minDelayMilliseconds;
+        private int _maxDelayMilliseconds;
+        private bool _hasRandomDelay;
+
+        private TimeSpan? _delay;
+
         /// <summary>
         /// The delay
         /// </summary>
```
StefH, line 36: Suggestion
Maybe better to remove this property.

And add two new properties:
- `MinimumDelay`
- `MaximumDelay`

And in case only the old situation is needed, just use one, like MinimumDelay.
And only calculate a random value is the MinimumDelay and MaximumDelay are defined

?

test/WireMock.Net.Tests/WireMockServerTests.cs
```diff
@@ -115,6 +115,43 @@ public async Task WireMockServer_Should_delay_responses_for_a_given_route()
...
+                await httClient.GetStringAsync("http://localhost:" + server.Ports[0] + "/foo");
+                return watch.ElapsedMilliseconds;
+            }
+
+            var elapsed1 = ExecuteTimedRequest();
+            var elapsed2 = ExecuteTimedRequest();
+            var elapsed3 = ExecuteTimedRequest();
+
+            // then
+            Check.That(elapsed1).IsNotEqualTo(elapsed2)
```
StefH, line 148: In theory it could be that all three values are the same ;-) Which makes you test break.

StefH (changes requested):
See comments.

And also add a test named `ResponseWithDelayTests` based on this class: ResponseWithBodyTests to test the new Delay code.

src/WireMock.Net/ResponseBuilders/Response.cs
```diff
@@ -295,6 +317,19 @@ public IResponseBuilder WithDelay(TimeSpan delay)
             return this;
         }
 
+        /// <inheritdoc cref="IDelayResponseBuilder.WithRandomDelay(int, int)"/>
+        public IResponseBuilder WithRandomDelay(int minimumMilliseconds = 0, int maximumMilliseconds = 60_000)
+        {
+            Check.Condition(minimumMilliseconds, d => d >= 0, nameof(minimumMilliseconds));
+            Check.Condition(maximumMilliseconds, d => d > 0, nameof(maximumMilliseconds));
```
StefH, line 324: You should also check that max > min

StefH (changes requested):

StefH:
@myarichuk Did you have time to check my comments?

StefH:
@myarichuk did you have time to check my code review comments?

StefH:
@myarichuk 
Thanks for this PR.

I'll need to add some changes, so I'll close this one and create a new PR myself.

PR #561, opened 2020-12-22
Do not save "admin" mappings when running in Proxy - mode

StefH:


StefH:
https://github.com/WireMock-Net/WireMock.Net/issues/559

PR #566, opened 2021-01-24
Do not save Mappings when SaveMappingForStatusCodePattern does not match

StefH:


StefH:
https://github.com/WireMock-Net/WireMock.Net/issues/565

PR #587, opened 2021-02-18
Fix WithCallback logic when using other fluent builder statements

StefH:


StefH:
https://github.com/WireMock-Net/WireMock.Net/issues/569

PR #597, opened 2021-03-21
Remove 2 second delay from first response and add IPv6 address support

### LINKED ISSUE #554: Request are very slow (~2.1 sec) - bad performance
### Describe the bug
Requests to the server are handled very slowly, ~2.1 sec for a simple get.

### Expected behavior:
Given that the server is in memory I would expect a response time of less de 50ms. 

### Test to reproduce
Simple add a stopwatch in any of the tests in https://github.com/WireMock-Net/WireMock.Net/blob/master/test/WireMock.Net.Tests/WireMockServerTests.cs 
I only count the request time, not the actual start of the server.
```
   var s = Stopwatch.StartNew();
            // then
            Check.That(server.Mappings).IsEmpty();
            Check.ThatAsyncCode(() => new HttpClient().GetStringAsync("http://localhost:" + server.Ports[0] + path)).ThrowsAny();
            s.Stop();
```
This gives 2.1 sec. I've measured the `WireMockMiddleware` and it takes 62ms. So something is lost on the way.

### Other related info
Using xunit for testing

benagain:
Fixes #554 

Remove an approximate two second delay in response to the first request from a new socket connection, apparently only occuring on some Windows 10 machines when using .Net Core.

A side-effect of this fix is that is also allows connections to IPv6 addresses.

### DIFF
--- a/src/WireMock.Net/Owin/AspNetCoreSelfHost.NETStandard.cs
+++ b/src/WireMock.Net/Owin/AspNetCoreSelfHost.NETStandard.cs
@@ -24,7 +24,7 @@ namespace WireMock.Owin
             {
                 if (urlDetail.IsHttps)
                 {
-                    kestrelOptions.Listen(System.Net.IPAddress.Any, urlDetail.Port, listenOptions =>
+                    kestrelOptions.ListenAnyIP(urlDetail.Port, listenOptions =>
                     {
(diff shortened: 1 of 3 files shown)


StefH:
@benagain Thank you.

benagain:
Thanks for merging so quickly @StefH!  It's made a great improvent to our test runs 😄

PR #675, opened 2021-11-01
Support basepath from servers

leolplex:
Hello @StefH ,

This change support the basepath definition in v2 and servers in  v3. The idea is when the file has a relative or absolute path this is attached at the path.

Thanks.

### DIFF
--- a/src/WireMock.Net.OpenApiParser/Mappers/OpenApiPathsMapper.cs
+++ b/src/WireMock.Net.OpenApiParser/Mappers/OpenApiPathsMapper.cs
@@ -234,17 +234,16 @@ namespace WireMock.Net.OpenApiParser.Mappers
         {
             if (servers == null || servers.Count == 0)
             {
-                return null;
+                return string.Empty;
             }
 
             OpenApiServer server = servers.First();
             Uri uriResult;
             if (Uri.TryCreate(server.Url, UriKind.RelativeOrAbsolute, out uriResult))
             {
-
                 return uriResult.IsAbsoluteUri ? uriResult.AbsolutePath : uriResult.ToString();
             }
-            return null;
+            return string.Empty;
         }
 
         private JToken MapOpenApiAnyToJToken(IOpenApiAny any)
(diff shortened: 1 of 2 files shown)


src/WireMock.Net.OpenApiParser/Mappers/OpenApiPathsMapper.cs
```diff
@@ -230,6 +230,23 @@ private string MapPathWithParameters(string path, IEnumerable<OpenApiParameter>
             return newPath;
         }
 
+        private string MapBasePath(IList<OpenApiServer> servers)
+        {
+            if (servers == null || servers.Count == 0)
+            {
+                return null;
```
StefH, line 237: Better to return `string.Empty` here?

src/WireMock.Net.OpenApiParser/Mappers/OpenApiPathsMapper.cs
```diff
@@ -230,6 +230,23 @@ private string MapPathWithParameters(string path, IEnumerable<OpenApiParameter>
...
+            }
+
+            OpenApiServer server = servers.First();
+            Uri uriResult;
+            if (Uri.TryCreate(server.Url, UriKind.RelativeOrAbsolute, out uriResult))
+            {
+
```
StefH, line 244: remove empty line

src/WireMock.Net.OpenApiParser/Mappers/OpenApiPathsMapper.cs
```diff
@@ -230,6 +230,23 @@ private string MapPathWithParameters(string path, IEnumerable<OpenApiParameter>
...
+                return uriResult.IsAbsoluteUri ? uriResult.AbsolutePath : uriResult.ToString();
+            }
+            return null;
```
StefH, line 247: Is this correct?

StefH (changes requested):

src/WireMock.Net.OpenApiParser/Mappers/OpenApiPathsMapper.cs
leolplex: Done!

src/WireMock.Net.OpenApiParser/Mappers/OpenApiPathsMapper.cs
leolplex: Done!

src/WireMock.Net.OpenApiParser/Mappers/OpenApiPathsMapper.cs
leolplex: Yes, if the Uril.TryCreate is false then a string.Empty should be returned. Comments?

StefH (approved):

PR #680, opened 2021-11-05
Support examples in properties

leolplex:
Hello @StefH ,

This PR contains the capability to support examples when it is defined in the properties:

```
"schemas": {
        "Order": {
          "type": "object",
          "properties": {
            "id": {
              "type": "integer",
              "format": "int64",
              "example": 10
            },
            "petId": {
              "type": "integer",
              "format": "int64",
              "example": 198772
            },
            "quantity": {
              "type": "integer",
              "format": "int32",
              "example": 7
            },
            "shipDate": {
              "type": "string",
              "format": "date-time"
            },
            "status": {
              "type": "string",
              "description": "Order Status",
              "example": "approved",
              "enum": [
                "placed",
                "approved",
                "delivered"
              ]
            },
            "complete": {
              "type": "boolean"
            }
          },
          "xml": {
            "name": "order"
          }
        },
}
```

The response looks like this:

```
 "Response": {
            "StatusCode": 200,
            "BodyAsJson": {
                "id": 10,
                "petId": 198772,
                "quantity": 7,
                "shipDate": "2022-08-09T23:23:26.490-05:00",
                "status": "approved",
                "complete": true
            },
            "Headers": {
                "Content-Type": "application/json"
            }
        }
```

It could be find in the examples that I uploaded before.

Thank you.

### DIFF
--- a/src/WireMock.Net.OpenApiParser/Utils/ExampleValueGenerator.cs
+++ b/src/WireMock.Net.OpenApiParser/Utils/ExampleValueGenerator.cs
@@ -26,33 +26,34 @@ namespace WireMock.Net.OpenApiParser.Utils
 
         public object GetExampleValue(OpenApiSchema schema)
         {
+            var schemaExample = schema?.Example;
             switch (schema?.GetSchemaType())
             {
                 case SchemaType.Boolean:
-                    OpenApiBoolean exampleBoolean = (OpenApiBoolean)schema?.Example;
-                    return exampleBoolean?.Value ?? _settings.ExampleValues.Boolean;
+                    var exampleBoolean = (OpenApiBoolean)schemaExample;
+                    return exampleBoolean is null ? _settings.ExampleValues.Boolean : exampleBoolean.Value;
 
                 case SchemaType.Integer:
                     switch (schema?.GetSchemaFormat())
                     {
                         case SchemaFormat.Int64:
-                            OpenApiLong exampleLong = (OpenApiLong)schema?.Example;
+                            var exampleLong = (OpenApiLong)schemaExample;
                             return exampleLong?.Value ?? _settings.ExampleValues.Integer;
+                        
                         default:
-                            OpenApiInteger exampleInteger = (OpenApiInteger)schema?.Example;
+                            var exampleInteger = (OpenApiInteger)schemaExample;
                             return exampleInteger?.Value ?? _settings.ExampleValues.Integer;
                     }
-
 
                 case SchemaType.Number:
                     switch (schema?.GetSchemaFormat())
                     {
                         case SchemaFormat.Float:
-                            OpenApiFloat exampleFloat = (OpenApiFloat)schema?.Example;
+                            var exampleFloat = (OpenApiFloat)schemaExample;
                             return exampleFloat?.Value ?? _settings.ExampleValues.Float;
 
                         default:
-                            OpenApiDouble exampleDouble = (OpenApiDouble)schema?.Example;
+                            var exampleDouble = (OpenApiDouble)schemaExample;
                             return exampleDouble?.Value ?? _settings.ExampleValues.Double;
                     }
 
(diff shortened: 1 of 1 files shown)


StefH:
Can you merge/rebase master into this branch ?

src/WireMock.Net.OpenApiParser/Utils/ExampleValueGenerator.cs
```diff
@@ -28,38 +29,55 @@ public object GetExampleValue(OpenApiSchema schema)
             switch (schema?.GetSchemaType())
             {
                 case SchemaType.Boolean:
-                    return _settings.ExampleValues.Boolean;
+                    OpenApiBoolean exampleBoolean = (OpenApiBoolean)schema?.Example;
```
StefH, line 32: You can just use `var` (multiple places in this file)

src/WireMock.Net.OpenApiParser/Utils/ExampleValueGenerator.cs
```diff
@@ -28,38 +29,55 @@ public object GetExampleValue(OpenApiSchema schema)
...
+                    return exampleBoolean?.Value ?? _settings.ExampleValues.Boolean;
 
                 case SchemaType.Integer:
-                    return _settings.ExampleValues.Integer;
+                    switch (schema?.GetSchemaFormat())
+                    {
+                        case SchemaFormat.Int64:
+                            OpenApiLong exampleLong = (OpenApiLong)schema?.Example;
+                            return exampleLong?.Value ?? _settings.ExampleValues.Integer;
+                        default:
```
StefH, line 41: add a newline

src/WireMock.Net.OpenApiParser/Utils/ExampleValueGenerator.cs
```diff
@@ -28,38 +29,55 @@ public object GetExampleValue(OpenApiSchema schema)
...
+                            OpenApiInteger exampleInteger = (OpenApiInteger)schema?.Example;
+                            return exampleInteger?.Value ?? _settings.ExampleValues.Integer;
+                    }
+
```
StefH, line 45: remove this extra newline

src/WireMock.Net.OpenApiParser/Utils/ExampleValueGenerator.cs
StefH, line 32: if you case to something (like OpenApiBoolean), the result cannot be null I think?

Not 100% sure....

Maybe I need to introduce NUllable and langversion 9+ in these projects to get better null checks..

StefH (changes requested):

src/WireMock.Net.OpenApiParser/Utils/ExampleValueGenerator.cs
leolplex: I changed the declaration type to var, the thing here is that `schema?.Example` is type `IOpenApiAny` then I need to force the cast to access the property `Value`.

src/WireMock.Net.OpenApiParser/Utils/ExampleValueGenerator.cs
leolplex: Done!

src/WireMock.Net.OpenApiParser/Utils/ExampleValueGenerator.cs
leolplex: Done!

src/WireMock.Net.OpenApiParser/Utils/ExampleValueGenerator.cs
leolplex: Yes you right, I changed the null validation to `exampleBoolean` because it could be null, not the value but the variable itself.

StefH (approved):

Issue #692, opened 2021-11-22
Case insensitive and ignoring optional path and header parameters in OpenApiPathsMapper


JanuszRybinski:
Hello @StefH and @leolplex,

related to (#691) but I decided to start a new discussion. While I testing OpenApi Parser, I encountered an interesting behavior of HttpRequestMessage, I will show on an example.

Assuming I have this definition of request:
```json
  "paths": {
    "/somethinks": {
      "put": {
        "parameters": [
          {
            "name": "X-Correlation-Id",
            "in": "header",
            "required": true,
            "schema": {
              "type": "string"
            }
          }
       ]
     }
  }
```

And I want to define a request for this:
```c#
using (var request = new System.Net.Http.HttpRequestMessage())
{
    request.Headers.TryAddWithoutValidation("X-Correlation-Id", "some value");
}
```

I did not check why exactly, but a header called "X-Correlation-ID" will be added to the request. By this small difference there is no matching mapping. In my opinion, headers should not be case-sensitive. To be sure, I checked what is in the RFC (https://www.rfc-editor.org/rfc/rfc7230#section-3.2). I have another idea to set case ignore in [MapRequestHeaders](https://github.com/WireMock-Net/WireMock.Net/blob/d9123ee3bdc800cfb0eb3fea19a58821abc8db23/src/WireMock.Net.OpenApiParser/Mappers/OpenApiPathsMapper.cs#L339):
```c#
var list = headers
    .Select(qp => new HeaderModel
    {
        Name = qp.Name,
        Matchers = new[]
        {
            GetExampleMatcherModel(qp.Schema, _settings.HeaderPatternToUse)
        },
        IgnoreCase = true
    })
    .ToList();
```
It seems to me that a similar setting also could be used in [MapQueryParameters](https://github.com/WireMock-Net/WireMock.Net/blob/d9123ee3bdc800cfb0eb3fea19a58821abc8db23/src/WireMock.Net.OpenApiParser/Mappers/OpenApiPathsMapper.cs#L323).

Additionally, I think that only the required parameters and headers should be taken into mappings:
```c#
var list = queryParameters
    .Where(qp => qp.Required)
```
```c#
var list = headers
    .Where(h => h.Required)
```

What is your opinion? If you agree, should I submit a new PR, or it can be added as part of #691?

Regards,
Janusz

leolplex:
Hello @JanuszRybinski / @StefH ,

[...]

PR #695, opened 2021-11-27
Allow configure IgnoreCase in settings

leolplex:
Hello @StefH  @JanuszRybinski,

This is the PR related with this #692, the idea is allow not case sensitive in Headers, Query Params, Request Body and examples.

Regards.

### DIFF
--- a/src/WireMock.Net.OpenApiParser/Settings/WireMockOpenApiParserSettings.cs
+++ b/src/WireMock.Net.OpenApiParser/Settings/WireMockOpenApiParserSettings.cs
@@ -36,5 +36,33 @@ namespace WireMock.Net.OpenApiParser.Settings
         /// Are examples generated dynamically?
         /// </summary>
         public bool DynamicExamples { get; set; } = false;
+
+        /// <summary>
+        /// Is headers case sensitive? (default is true).        
+        /// true = no case sensitive;
+        /// false = case sensitive 
+        /// </summary>
+        public bool IgnoreCaseHeaders { get; set; } = true;
+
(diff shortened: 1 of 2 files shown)


StefH (changes requested):
It's better to move this setting (**IgnoreCase**) to `WireMockOpenApiParserSettings.cs` to that it can be modified by user.

Default value can be set to `true`

leolplex:
@StefH,

you're right it's better idea, done!

src/WireMock.Net.OpenApiParser/Settings/WireMockOpenApiParserSettings.cs
```diff
@@ -36,5 +36,33 @@ public class WireMockOpenApiParserSettings
         /// Are examples generated dynamically?
         /// </summary>
         public bool DynamicExamples { get; set; } = false;
+
+        /// <summary>
+        /// Is headers case sensitive? (default is true).        
+        /// true = no case sensitive;
```
StefH, line 42: No need to add these two lines:
```
        /// true = no case sensitive;
        /// false = case sensitive 
```

You can remove these.

StefH (changes requested):

src/WireMock.Net.OpenApiParser/Settings/WireMockOpenApiParserSettings.cs
leolplex: @StefH  Done!

StefH (approved):

Issue #685, opened 2021-11-17
GuidWildcardMatcher to match on GUIDs


brogdogg:
**Is your feature request related to a problem? Please describe.**

I have been using the `RegexMatcher` to do matches for paths with GUIDs, but after a while the `Pattern` in the JSON mappings becomes unwieldly long in length when a few GUIDs are used.

**Describe the solution you'd like**

Would like a matcher to allow for matching a GUID (mostly in the `Path`). For example:

- `Path` -  `http://localhost/v1.0/4a7eaad5-bbe7-4579-bb0b-7c85352fe1b6/task/edef5b4a-33bb-4920-bb42-1d5e6283984e`
- JSON mapping:

  ```JSON
  {
      "Request" : 
      {
        "Path": {
          "Matchers": [{
            "Name": "GuidWildcardMatcher",
            "Pattern": "/*/{D}/task/{D}",
            "IgnoreCase": true
          }]
        }
      }
  }
  ```

Where the tokens used in the pattern match up with the `GUID.ToString(string? format)` method as described [here](https://docs.microsoft.com/en-us/dotnet/api/system.guid.tostring?view=net-5.0#System_Guid_ToString_System_String_).

**Describe alternatives you've considered**

None - I can use the `RegexMatcher` if the feature is not accepted.

**Is your feature request supported by [WireMock (java version)](https://www.wiremock.org)? Please provide details.**

I do not believe this is supported in the java version (I did a search and couldn't find anything that I could see)

StefH:
It would also be an option to update the wildcard matcher.
Currently the wildcard matcher supports a `*`. Maybe when adding logic to understand also `GUID(D)` could be an option...

Just thinking....

brogdogg:
True, I mean would it be a more general matcher available in JSON as well?  I forked and trying to play with it to see.

brogdogg:
OK, after playing I see what you mean by updating the wildcard matcher, since it extends `RegexMatcher`.  I was trying something like this:

```c#
            return patterns.Select(pattern => new AnyOf<string, StringPattern>(
                new StringPattern
                {
                    Pattern = "^" + Regex.Escape(pattern.GetPattern())
                                         .Replace(@"\{GUIDD}", "([A-Z0-9]{8}-([A-Z0-9]{4}-){3}[A-Z0-9]{12})")
                                         .Replace(@"\{GUIDN}", "([A-Z0-9]{32})")
[...]

PR #699, opened 2021-12-08
GUID Pattern support in RegexMatcher

brogdogg:
Adding a new `RegexGuid` class, extending the `System.Text.RegularExpressions.Regex` class, to support patterns for identifying a GUID with respect to the way the [System.Guid.ToString()](https://docs.microsoft.com/en-us/dotnet/api/system.guid.tostring?view=net-6.0) formats a new GUID.

Additionally, the `RegexMatcher` was updated to use this new `RegexGuid` class.

Potential use:

```c#
// Create a regex matcher with a \GUIDD pattern (i.e. only match on all caps GUID with format D
var matcher = new RegexMatcher("\GUIDD");
// Would not match, since there are lower case
double result = matcher.IsMatch("5ae7dbb5-ea2c-4c55-ac38-9634681f168d");
// Would match, as it is a valid GUID in all upper case and valid with respect o the
// format D specifer.
result = matcher.IsMatch("5AE7DBB5-EA2C-4C55-AC38-9634681F168D");
```

This is a suggested change to address feedback to #685


brogdogg:
Status check shows a failed unit test.  The test that failed I did not touch and when I run all tests locally I am unable to recreate the failure.

StefH:
Sometimes that test fails... I did update that test just now 

I did rerun your CI build and it is ok 

Maybe you can merge latest master into your PR branch?

src/WireMock.Net/RegularExpressions/RegexGuid.cs
```diff
@@ -0,0 +1,167 @@
...
+    /// Extension to the <see cref="Regex"/> object, adding support for GUID
+    /// tokens for matching on.
+    /// </summary>
+    public class RegexGuid : Regex
+    {
+        /// <summary>
+        /// Token for a GUID formated with `B` format specifier with lower case
+        /// values.
+        /// </summary>
+        public const string GuidBLowerToken = @"\guidb";
```
StefH, line 20: An easier and (smaller code) approach on this would be to define a dictionary with the tokens and regex.
Like:
``` c#
private static readonly IDictionary<string,string> _replacements = new Dictionary<string, string>
{
 { @"\guidb",  @"(\{[a-z0-9]{8}-([a-z0-9]{4}-){3}[a-z0-9]{12}\})" }, // Regular expression pattern associated with the expected format for `B` format specifier with lower case.
 // . . .
}
```

src/WireMock.Net/RegularExpressions/RegexGuid.cs
```diff
@@ -0,0 +1,167 @@
+using System;
+using System.Collections.Generic;
+using System.Linq;
+using System.Text;
+using System.Text.RegularExpressions;
+using System.Threading.Tasks;
+
+namespace WireMock.RegularExpressions
```
StefH, line 8: General: you can just write comments / code on 1 line, no need to break at 80 characters.

src/WireMock.Net/RegularExpressions/RegexGuid.cs
```diff
@@ -0,0 +1,167 @@
...
+        }
+
+        /// <summary>
+        /// Replaces all instances of valid GUID tokens with the correct regular
+        /// expression to match.
+        /// </summary>
+        /// <param name="pattern">
+        /// Pattern to replace token for.
+        /// </param>
+        private static string ReplaceGuidPattern(string pattern)
```
StefH, line 155: When using the dictionary, this can be simpler.
And a null check should be added.

``` c#
private static string ReplaceGuidPattern(string pattern)
{
    Check.NotNull(pattern, nameof(pattern));

     foreach (var replacement in _replacements)
     {
         pattern = pattern.Replace(replacement.Key, replacement.Value);
     }
}


```

StefH (changes requested):
Thanks a lot.

1] take a look at my comments

2] I did not check all regex, but is there also a simple GUID match which allows any casing? Like `AAAAa1ff-2f1e-4f44-a053-e2c34a7ccBBB`?

3] Maybe it's good to replace all existing Regex usages in WireMock with this new one.

4] I'm thinking if we need to make using this new Regex configurable via the wiremock settings maybe?

brogdogg:
> 2] I did not check all regex, but is there also a simple GUID match which allows any casing? Like AAAAa1ff-2f1e-4f44-a053-e2c34a7ccBBB?

I thought about this, but since it is a `Regex` the ignore case operator could be used, wouldn't you think?

> 3] Maybe it's good to replace all existing Regex usages in WireMock with this new one.

Are you requesting I change all references to the `RegexGuid` class?

> 4] I'm thinking if we need to make using this new Regex configurable via the wiremock settings maybe?

I'm not sure what you mean here, would you be able to point to a sample?  (Sorry, I'm still new to Wiremock and have only used JSON files for configuring the dotnet tool).

StefH:
2]
You are correct

3 and 4]
Maybe better to rename `RegexGuid` to `RegexExtended` because in the future. more replacements can be added.

And maybe add an property like `bool UseExtendedRegex` to the settings:
- https://github.com/WireMock-Net/WireMock.Net/blob/master/src/WireMock.Net.Abstractions/Admin/Settings/SettingsModel.cs
- https://github.com/WireMock-Net/WireMock.Net/blob/master/src/WireMock.Net/Settings/WireMockServerSettings.cs
- https://github.com/WireMock-Net/WireMock.Net/blob/master/src/WireMock.Net/Settings/WireMockServerSettingsParser.cs

And the default value is set to true, but if needed, a user can set this to false.

StefH:
@brogdogg 
If you have time, can apply my comments?
Except for 3 and 4, I can also take that up once your branch is merged into master.

brogdogg:
> @brogdogg If you have time, can apply my comments? Except for 3 and 4, I can also take that up once your branch is merged into master.

Hey @StefH yeah, I was planning to address this weekend.  Current job is calling for heavy hours right now.  Sorry, but will look into it.

StefH (approved):

PR #706, opened 2021-12-23
Provide open api schema to dynamic examples generator so you can generate accurate data

brunotarghetta:
@StefH  Having ope api Schema we can use settings like max-length in case of a string to generate more accurate mock data when you implement your dynamic examples

### DIFF
--- a/examples/WireMock.Net.OpenApiParser.ConsoleApp/DynamicDataGeneration.cs
+++ b/examples/WireMock.Net.OpenApiParser.ConsoleApp/DynamicDataGeneration.cs
@@ -1,0 +1,48 @@
+using System;
+using Microsoft.OpenApi.Models;
+using RandomDataGenerator.FieldOptions;
+using RandomDataGenerator.Randomizers;
+using WireMock.Net.OpenApiParser.Settings;
+
+namespace WireMock.Net.OpenApiParser.ConsoleApp
+{
+
+    public class DynamicDataGeneration : IWireMockOpenApiParserExampleValues
+    {
+
+        public bool Boolean { get { return RandomizerFactory.GetRandomizer(new FieldOptionsBoolean()).Generate() ?? true; } set { } }
+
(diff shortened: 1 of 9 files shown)


StefH (changes requested):
See my comments.

And I wonder how to can be used, can you add an example (maybe in the https://github.com/WireMock-Net/WireMock.Net/tree/master/examples/WireMock.Net.OpenApiParser.ConsoleApp ?)

brunotarghetta:
@StefH I've made the changes you suggested. Also I added some code on the ConsoleApp to show you how this can be useful.
Let me share some images

In you swagger file you can specify some attributes (maxLength on string types)
Screen Shot 2021-12-27 at 14 21 46

Screen Shot 2021-12-27 at 14 21 24

So, if you are working with your own example values generator, you can use this settings

Screen Shot 2021-12-27 at 14 41 43

as a result you can check on mapping those examples string with their own length


Screen Shot 2021-12-27 at 14 36 26

In our case this is really useful

Let me know if you need more information.

Thanks in advance.

StefH (changes requested):
Looks good.

Can you do these 2 changes:

Add `virtual` to all public methods in:
- `WireMockOpenApiParserExampleValues.cs`
- `WireMockOpenApiParserDynamicExampleValues.cs`

With that change, you can just easily extend from the existing class and override the existing methods. (Can be useful for your new file "DynamicDataGeneration.cs").

brunotarghetta:
@StefH I've made virtual changes and I extend my DynamicDataGeneration.

StefH (approved):

brunotarghetta:
Hi @StefH, good morning! do you know when this features will be release?

StefH:
@brunotarghetta 

New NuGet packages will be released in a few minutes...

brunotarghetta:
> @brunotarghetta
> 
> New NuGet packages will be released in a few minutes...

That sound great!!!!!!! Thanks!!

PR #736, opened 2022-03-09
Remove interface for all Setting classes

StefH:


PrzemekScott:
Solution is working as expected - all settings for proxying are loaded properly, when set in appsettings.json file, at the start of the app. 

image

StefH:
@PrzemekScott 
A new version will be released to NuGet tomorrow.

PR #759, opened 2022-05-20
Add WireMock.Net.xUnit project

StefH:


src/WireMock.Net.xunit/TestOutputHelperWireMockLogger.cs
```diff
@@ -0,0 +1,79 @@
+using System;
+using Newtonsoft.Json;
+using Stef.Validation;
+using WireMock.Admin.Requests;
+using WireMock.Logging;
+using Xunit.Abstractions;
+
+namespace WireMock.Net.xunit;
```
siewers, line 8: I think the namespace should be `WireMock.Net.Xunit` to follow the same naming as the xUnit project.

siewers:
I guess folders, assembly names etc. should be renamed to `Xunit` as well :)

StefH:
I was doubting about the name.

I see some NuGet packages like `xxx.xUnit` and I see `xxx.XUnit` and `xxx.Xunit`
image


And xunit itself is named:
image


So what to do ?

siewers:
Yes, it's a mess.
I would suggest to stick with the same namespace as used in the xUnit framework.

I would say in namespaces and assembly use `Xunit`.
In all documentation and text use `xUnit`.

The all lowercase version is not really used anywhere besides the xUnit assemblies, which is likely of legacy reasons.

It's basically caused by the conflict of C# using pascal case but xUnit starting with a lowercase letter which makes it a bit difficult 🙂

PR #797, opened 2022-08-22
Fix WithHeader when using RejectOnMatch

### LINKED ISSUE #796: RequestMessageHeaderMatcher with MatchBehaviour.RejectOnMatch reverses match results twice
### Describe the bug
Following the [Reversing the match behaviour with MatchBehaviour.RejectOnMatch
](https://github.com/WireMock-Net/WireMock.Net/wiki/Request-Matching#reversing-the-match-behaviour-with-matchbehaviourrejectonmatch) example the reponse will always be `Unauthorized` even if a correct header is added on the request.

I believe the following line shouldn't call `MatchBehaviourHelper.Convert()` again as it is already called for each Matcher. Other RequestMessage***Matcher's do not call `MatchBehaviourHelper.Convert()` before returning.
https://github.com/WireMock-Net/WireMock.Net/blob/be4b0addca8d89cedbdbbd645c842417a04b8178/src/WireMock.Net/Matchers/Request/RequestMessageHeaderMatcher.cs#L138

### Expected behavior:
Given the following example taken from the docs. The expected behavior is that only an `Unauthorized` response is returned when the header `api-key` is missing from the http request.

``` C#
server
   .Given(Request.Create()
             .WithPath("/needs-a-key")
             .UsingGet()
             .WithHeader("api-key", "*", MatchBehaviour.RejectOnMatch)
             .UsingAnyMethod())
   .RespondWith(Response.Create()
             .WithStatusCode(HttpStatusCode.Unauthorized)
             .WithBody(@"{ ""result"": ""api-key missing""}"));
```

### Other related info
Since WireMock.Net v1.5.0.

flts:
Fixes #796

### DIFF
--- a/src/WireMock.Net/Matchers/Request/RequestMessageHeaderMatcher.cs
+++ b/src/WireMock.Net/Matchers/Request/RequestMessageHeaderMatcher.cs
@@ -135,7 +135,7 @@ public class RequestMessageHeaderMatcher : IRequestMatcher
                 results.Add(MatchScores.ToScore(resultsPerMatcher, MatchOperator.And));
             }
 
-            return MatchBehaviourHelper.Convert(_matchBehaviour, MatchScores.ToScore(results, MatchOperator));
+            return MatchScores.ToScore(results, MatchOperator);
         }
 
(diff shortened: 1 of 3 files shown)


StefH:
@flts Thank you.

PR #798, opened 2022-08-27
Add support to use 'mapping' object in in reponse templating

StefH:


StefH:
https://github.com/WireMock-Net/WireMock.Net/issues/772

PR #802, opened 2022-09-02
Add assertions for request methods

rafaelmfonseca:
Hi, i've added some assertions for request methods.

### DIFF
--- a/src/WireMock.Net.FluentAssertions/Assertions/WireMockAssertions.cs
+++ b/src/WireMock.Net.FluentAssertions/Assertions/WireMockAssertions.cs
@@ -129,4 +129,59 @@ public class WireMockAssertions
 
         return new AndConstraint<WireMockAssertions>(this);
     }
+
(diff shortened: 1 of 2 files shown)


StefH:
@rafaelmfonseca 
Thank you,

However, one test fails:
 Failed WireMock.Net.Tests.FluentAssertions.WireMockAssertionsTests.HaveReceivedACall_UsingConnect_WhenACallWasMadeUsingConnect_Should_BeOK [6 ms]
  Error Message:
   System.Net.Http.HttpRequestException : CONNECT request must contain Host header.

rafaelmfonseca:
@StefH
Thanks!
I've fixed the `UsingConnect()` test, except for net452, it's like HttpClient is not sending the request using the method "CONNECT"... I tried a lot of thing but it didn't work. It's weird because with NET461+ it's working, so I assume some internal code of HttpClient changed from NET452 to NET461. Do you have any ideas?

PR #871, opened 2023-01-05
Add UseWebhooksFireAndForget to Server ConvertMapping

ggradnig:
Hi,

this mapping was missing, so when using the Admin API the UseWebhooksFireAndForget flag was always false.


StefH:
Thanks.

PR #918, opened 2023-04-05
Change some internal classes to public

Naudate:
Hello, 

I am using WireMock.Net in a project and I wanted to add some changes and see with you if it was possible. 
We are using WireMock.Net to mock requests and reuse mappings for JMeter tests.  I can't use the recording process that is already in place because its a proxy . We prefer that the application make the call rather than WireMock.Net. So to use what have been done and stay up to date I wanted to change some classes :
- **_JsonSerializationConstants_**
- **_BodyParser_**
- _**BodyParserSettings**_
Those classes are internal and I wanted to put them public. That way I can use WireMock.Net in the project.

Thank you for your answer.

### DIFF
--- a/src/WireMock.Net/Serialization/JsonSerializationConstants.cs
+++ b/src/WireMock.Net/Serialization/JsonSerializationConstants.cs
@@ -3,7 +3,7 @@ using Newtonsoft.Json.Serialization;
 
 namespace WireMock.Serialization;
 
-internal static class JsonSerializationConstants
+public static class JsonSerializationConstants
 {
     public static readonly JsonSerializerSettings JsonSerializerSettingsDefault = new()
     {
--- a/src/WireMock.Net/Util/BodyParser.cs
+++ b/src/WireMock.Net/Util/BodyParser.cs
@@ -11,7 +11,7 @@ using WireMock.Types;
 
 namespace WireMock.Util;
 
-internal static class BodyParser
+public static class BodyParser
 {
     private static readonly Encoding DefaultEncoding = new UTF8Encoding(encoderShouldEmitUTF8Identifier: false, throwOnInvalidBytes: true);
     private static readonly Encoding[] SupportedBodyAsStringEncodingForMultipart = { DefaultEncoding, Encoding.ASCII };
--- a/src/WireMock.Net/Util/BodyParserSettings.cs
+++ b/src/WireMock.Net/Util/BodyParserSettings.cs
@@ -2,7 +2,7 @@ using System.IO;
 
 namespace WireMock.Util;
 
-internal class BodyParserSettings
+public class BodyParserSettings
 {
     public Stream Stream { get; set; } = null!;
 


StefH:
Can't you just use the MappingModels for this?

Naudate:
I am using MappingModel to create the mapping structure. But in order to create my mappings using json I need those three classes.

StefH:
Did you try using the `IMappingBuilder` / `MappingBuilder` ?

Naudate:
Yes I tried. My issue is that I have a the request in this format `Wiremock.RequestMessage` and the response in this format  `Wiremock.ResponseMessage`. I put them in a `ToMapping` where I create the mapping structure and in order to get the body type I'm using `BodyParser.DetectBodyTypeFromContentType(string? contentTypeValue)` which in an internal that I can't use. This the first blocker and then after I have a ToJson class where i put my MappingModel from `ToMapping` and serialize it to Json like you did in [MappingBuilder.cs](https://github.com/WireMock-Net/WireMock.Net/blob/master/src/WireMock.Net/MappingBuilder.cs#LL156C5-L159C6). maybe there is something I'm missing let know

Naudate:
Hello StefH, 

I don't see how else to do it.

StefH:
@Naudate 
I still don't get what you are trying to do.

Can you please give a code - example on what you are trying to achieve?

Naudate:
Hello @StefH,

Yes, I wasn't clear about what I'm trying to do.
I'm working on an app in microservices where these services send http request to external services.
We have an issue where in our JMeter integretion test some external services changes things that they are sending and those changes make the JMeter test fail. So we are making a mockservice where we can record the external calls and replay them.

First I when with the proxy and record method that you provide but wasn't ideal because we couldn't record the call. We couldn't do that because we are using [kerberos authentification](https://learn.microsoft.com/en-us/windows-server/security/kerberos/kerberos-authentication-overview) so if wiremock make the call it doesn't work because it is the app that identify when microservice are use. So using proxy doesn't work. That is why we headed using wiremock our own recording process. Here is little vue of what I'm doing.
```c#
protected override async Task<HttpResponseMessage> SendAsync(
     HttpRequestMessage request,
     CancellationToken cancellationToken)
{
    bool testing = GetTestingFromHeaders(request);
    bool recording = GetRecordingFromHeaders(request);
    if(recording)
    {
        response = await base.SendAsync(request, cancellationToken);
        StringBuilder sb = CreateMappingName(request);
        IRequestMessage req = MapToRequestMessage(request);
        ResponseMessage res = MapToResponseMessage(response);

        MappingModel newmapping = ToMapping(req, res, sb.ToString());
        string input = ToJson(newmapping);
        var c = new StringContent(input, System.Text.Encoding.UTF8);

        maprequest.Method = HttpMethod.Post;
        maprequest.RequestUri = new Uri(
            $"{httpMocksServiceUri.Value}/__admin/files/{sb}.json");
        maprequest.Content = c;
        return await base.SendAsync(maprequest, cancellationToken);

    }
    else if(testing)
    {
        request.RequestUri = new Uri(
        $"{httpMocksServiceUri}{request.RequestUri.PathAndQuery}");
        response = await base.SendAsync(request, cancellationToken);
        return response;
    }
    return await base.SendAsync(request, cancellationToken);
}
public string ToJson(MappingModel newmapping)
{
    //Where I need src/WireMock.Net/Serialization/JsonSerializationConstants.cs to be public 
    return JsonConvert.SerializeObject(newmapping, JsonSerializationConstants.JsonSerializerSettingsDefault);
}

public ResponseMessage MapToResponseMessage(HttpResponseMessage response)
{
    var responseMessage = new ResponseMessage();
    var bodydata = new BodyData();
    BodyType bodytype = new BodyType();
    if (response.Content.Headers.ContentType != null)
    {
        //Where i need src/WireMock.Net/Util/BodyParser.cs to be public
        bodytype = BodyParser.DetectBodyTypeFromContentType(response.Content.Headers.ContentType.MediaType);
    }
    else
    {
        bodytype = BodyType.None;
    }
    foreach (var header in response.Headers)
    {
        responseMessage.Headers.Add(header.Key, header.Value.First());
    }
    switch (bodytype)
    {
        case BodyType.String:
            var bodys = response.Content.ReadAsStringAsync().Result;
            bodydata = new BodyData()
            {
                BodyAsString = bodys
            };
            break;
        case BodyType.Json:
            var bodyj = response.Content.ReadAsStringAsync().Result;
            bodydata = new BodyData()
            {
                BodyAsJson = bodyj
            };
            break;
        case BodyType.Bytes:
            var bodyb = response.Content.ReadAsByteArrayAsync().Result;
            bodydata = new BodyData()
            {
                BodyAsBytes = bodyb
            };
            break;
        default:
            break;
    }
    responseMessage.StatusCode = response.StatusCode;
    responseMessage.BodyData = bodydata;
    responseMessage.BodyDestination = BodyDestinationFormat.SameAsSource;
    return responseMessage;
}
``` 

SendAsync is heading to wiremock standalone app if we are recording or testing :
```c#
MocksServer = StandAloneApp.Start(new WireMockServerSettings
{
    MaxRequestLogCount = int.Parse(_configuration.GetSection("MaxRequestLogCount").Value),
    Urls = _configuration.GetSection("ListeningUrls").Value.Split(','),
    StartAdminInterface = true,
    ReadStaticMappings = true,
    WatchStaticMappings = true,
    Logger = new WireMockLogger(_logger)
});
``` 
`ToMapping` is like https://github.com/WireMock-Net/WireMock.Net/blob/master/src/WireMock.Net/Serialization/ProxyMappingConverter.cs#L29
But i had to change it because I don't put the same values in it.

That is it I think. You should have all of the info that I can give you. let me know if something isn't clear.

Thank You

StefH:
Just a quick remark, that `MapToResponseMessage` method, looks a lot like this: https://github.com/WireMock-Net/WireMock.Net/blob/master/src/WireMock.Net/Http/HttpResponseMessageHelper.cs#L13 which is used in https://github.com/WireMock-Net/WireMock.Net/blob/master/src/WireMock.Net/Proxy/ProxyHelper.cs#L40

Naudate:
Hello @StefH,
I close the pull request to come up with a better solution.

PR #926, opened 2023-05-02
Fix C# mapping code generator for header names

cezarypiatek:
This PR removes redundant ")" character from header names.


StefH (approved):

StefH:
@cezarypiatek 
Thank you

PR #933, opened 2023-05-07
C# code generator improvements

cezarypiatek:
This PR brings the following improvements:
- Add quote escaping in generated C# strings
- Handle JSON in the response body. Json is converted to C# anonymous object definition and fed into `WithBodyAsJson` method.

### DIFF
--- a/src/WireMock.Net/Serialization/MappingConverter.cs
+++ b/src/WireMock.Net/Serialization/MappingConverter.cs
@@ -5,6 +5,7 @@ using System.Net;
 using System.Text;
 using System.Threading;
 using Newtonsoft.Json;
+using Newtonsoft.Json.Linq;
 using Stef.Validation;
 using WireMock.Admin.Mappings;
 using WireMock.Constants;
@@ -143,13 +144,26 @@ internal class MappingConverter
             }
         }
 
-        if (response.ResponseMessage.BodyData is { })
+        if (response.ResponseMessage.BodyData is { } bodyData)
         {
             switch (response.ResponseMessage.BodyData.DetectedBodyType)
             {
                 case BodyType.String:
                 case BodyType.FormUrlEncoded:
                     sb.AppendLine($"        .WithBody(\"{EscapeCSharpString(bodyData.BodyAsString)}\")");
+                    break;
+                case BodyType.Json:
+                    if (bodyData.BodyAsJson is string bodyStringValue)
+                    {
+                        sb.AppendLine($"        .WithBody(\"{EscapeCSharpString(bodyStringValue)}\")");
+                    }
+                    else
+                    {
+                        var serializedBody = JsonConvert.SerializeObject(bodyData.BodyAsJson);
+                        var deserializedBody = JToken.Parse(serializedBody);
+                        sb.AppendLine($"        .WithBodyAsJson({ConvertJsonToAnonymousObjectDefinition(deserializedBody, 2)})");
+                    }
+
                     break;
             }
         }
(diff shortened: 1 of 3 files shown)


src/WireMock.Net/Serialization/MappingConverter.cs
```diff
@@ -173,6 +187,49 @@ public string ToCSharpCode(IMapping mapping, MappingConverterSettings? settings
         return sb.ToString();
     }
 
+    private string ConvertJsonToAnonymousObjectDefinition(JToken token, int ind = 0)
+    {
+        string FormatObject(JObject jObject)
+        {
+            var indStr = new string(' ', 4 * ind);
+            var indStrSub = new string(' ', 4 * (ind+1));
```
StefH, line 195: can you do a format document in visual studio?

src/WireMock.Net/Serialization/MappingConverter.cs
```diff
@@ -173,6 +187,49 @@ public string ToCSharpCode(IMapping mapping, MappingConverterSettings? settings
         return sb.ToString();
     }
 
+    private string ConvertJsonToAnonymousObjectDefinition(JToken token, int ind = 0)
```
StefH, line 190: Another possible option would be to just use reflection:
``` c#
private static string GenerateObjectInitializationCodeRecursive(object anonymousObject)
{
	var properties = anonymousObject.GetType().GetProperties();
	var objectBuilder = new StringBuilder();

	objectBuilder.AppendLine("new");
	objectBuilder.AppendLine("{");

	foreach (var property in properties)
	{
		string propertyName = property.Name;
		object propertyValue = property.GetValue(anonymousObject);

		if (propertyValue != null)
		{
			if (propertyValue is string)
			{
				objectBuilder.AppendLine($"    {propertyName} = \"{propertyValue}\",");
			}
			else if (property.PropertyType.IsClass && !property.PropertyType.FullName.StartsWith("System."))
			{
				string nestedObjectCode = GenerateObjectInitializationCodeRecursive(propertyValue);
				objectBuilder.AppendLine($"    {propertyName} = {nestedObjectCode.Trim()},");
			}
			else
			{
				objectBuilder.AppendLine($"    {propertyName} = {propertyValue},");
			}
		}
	}

	objectBuilder.AppendLine("};");

	return objectBuilder.ToString();
}
```

StefH (changes requested):

src/WireMock.Net/Serialization/MappingConverter.cs
cezarypiatek: Fixed

src/WireMock.Net/Serialization/MappingConverter.cs
cezarypiatek: Yes, it's quite common that a given problem can be solved in multiple ways. Try to use the code from above and there are compilation issues (probably due to missing properties in some target frameworks) and I guess some additional adjustments are needed to achieve proper formatting.

Is there any reason why Newtonsoft.Json can't be used here? It's already referenced by WireMock project.

StefH: Your solution is fine. It was just an idea.

StefH (approved):

cezarypiatek:
Thanks for merging. Btw, do you have a plan for the next release? I would like to start using the code generation feature but I need those recent changes.

StefH:
@cezarypiatek Today I'll release new NuGet.

PR #934, opened 2023-05-08
Code generator improvements

cezarypiatek:


cezarypiatek:
@StefH would you mind reviewing/merging this pr? Thanks in advance.

StefH:
> @StefH would you mind reviewing/merging this pr? Thanks in advance.

will do later today

src/WireMock.Net/Serialization/MappingConverter.cs
```diff
@@ -537,4 +555,85 @@ private static string FormatArray(JArray jArray, int ind)
 
         return $"new [] {{ {string.Join(", ", items)} }}";
     }
+
+    private static readonly HashSet<string> csharp_keywords = new HashSet<string>(new[]
```
StefH, line 559: Can you move this to the top of the file?

(And maybe instead of using this list, you can maybe use:
https://github.com/StefH/ProxyInterfaceSourceGenerator/blob/main/src/ProxyInterfaceSourceGenerator/Extensions/SymbolExtensions.cs#L35
) ?

StefH (changes requested):
Some small question... See comments

src/WireMock.Net/Serialization/MappingConverter.cs
cezarypiatek: Using `SyntaxFacts.GetKeywordKind` requires reference to `Microsoft.CodeAnalysis.CSharp`.
I can extract the whole logic responsible for generating anonymous object to a separate class, what fo you think?

StefH: Yes, maybe a new internal static class, placed in `Util` folder.

cezarypiatek: fixed

StefH: Thanks

StefH (approved):

PR #1043, opened 2023-12-16
FindRequestByMappingGuidAsync

StefH:


src/WireMock.Net.RestClient/IWireMockAdminApi.cs
```diff
@@ -192,14 +192,22 @@ public interface IWireMockAdminApi
...
     [Header("Content-Type", "application/json")]
     Task<IList<LogEntryModel>> FindRequestsAsync([Body] RequestModel model, CancellationToken cancellationToken = default);
 
+    /// <summary>
+    /// Find a request based on the Mapping Guid.
+    /// </summary>
+    /// <param name="mappingGuid">The Mapping Guid</param>
+    /// <param name="cancellationToken">The optional cancellationToken.</param>
+    [Get("requests/find")]
+    Task<LogEntryModel?> FindRequestByMappingGuidAsync([Query] Guid mappingGuid, CancellationToken cancellationToken = default);
```
tlevesque-ueat, line 209: It should probably be `FindRequestsByMappingGuidAsync`, and return a collection

src/WireMock.Net/Server/WireMockServer.Admin.cs
```diff
@@ -600,6 +601,26 @@ private IResponseMessage RequestsFind(IRequestMessage requestMessage)
...
     }
+
+    private IResponseMessage RequestFindByMappingGuid(IRequestMessage requestMessage)
+    {
+        if (requestMessage.Query != null &&
+            requestMessage.Query.TryGetValue("mappingGuid", out var value) &&
+            Guid.TryParse(value.ToString(), out var mappingGuid)
+        )
+        {
+            var logEntry = LogEntries.SingleOrDefault(le => !le.RequestMessage.Path.StartsWith("/__admin/") && le.MappingGuid == mappingGuid);
```
tlevesque-ueat, line 612: There could be more than one matching request, and in this case `SingleOrDefault` will throw.

src/WireMock.Net/Server/WireMockServer.Admin.cs
```diff
@@ -600,6 +601,26 @@ private IResponseMessage RequestsFind(IRequestMessage requestMessage)
...
+            if (logEntry != null)
+            {
+                var logEntryMapper = new LogEntryMapper(_options);
+                return ToJson(logEntryMapper.Map(logEntry));
+            }
+
+            return ResponseMessageBuilder.Create(HttpStatusCode.OK);
```
tlevesque-ueat, line 619: Returning 200 with an empty response is probably not what the client would expect. If the status code is 200, the client will expect the body to contain a `LogEntryModel`.

But anyway, I think this endpoint should return a collection (which will be empty if no match is found)

tlevesque-ueat (commented):
This review comes a bit late, sorry about that... I didn't pay enough attention when you first sent the PR.
I realize changing it now would be a breaking change, but I think it's worth it, because at the moment it's broken (it will throw if there's more than one matching request)

PR #1085, opened 2024-03-19
Fix FluentAssertions (actual body is not displayed in error message)

### LINKED ISSUE #1084: FluentAssertions - Actual body is not displayed in error message when using Json Body
### Describe the bug
When using a Json Body, Fluent assertion cannot format the actual body in the error message

> Expected wiremockserver to have been called using body { param = value2 }, but didn't find it among the body {{{{empty}}}}.

### Expected behavior:

The json body is correctly formatted

> Expected wiremockserver to have been called using body { param = value2 }, but didn't find it among the body {{
  "param": "value"
}}.

### Test to reproduce

You can reproduce the message with this unit test (proper assertion is not done, the test fails with the unformatted message)

```cs
    [Fact]
    public async Task HaveReceived1Call_WithBodyAsJson_UsingJsonMatcher()
    {
        // Arrange
        var server = WireMockServer.Start();

        server
            .Given(Request.Create().WithPath("/a").UsingPost())
            .RespondWith(Response.Create().WithBody("A response"));

        // Act
        var httpClient = new HttpClient();

        var body = new
        {
            param = "value"
        };
        await httpClient.PostAsJsonAsync($"{server.Url}/a", body);

        body = new
        {
            param = "value2"
        };
        // Assert
        server
            .Should()
            .HaveReceived(1)
            .Calls()
            .WithBodyAsJson(new JsonMatcher(body))
            .And
            .UsingPost();

        server.Stop();
    }
```

### Other related info

One solution is to implement a custom `IValueFormatter`

Example:
```cs
public class JObjectFormatter : IValueFormatter
{
    public bool CanHandle(object value)
    {
        return value is JObject;
    }

    public void Format(object value, FormattedObjectGraph formattedGraph, FormattingContext context, FormatChild formatChild)
    {
        var jObject = (JObject)value;
        formattedGraph.AddFragment(jObject.ToString());
    }
}
```

And to add it in the custom formatters `Formatter.AddFormatter(new JObjectFormatter());`

Note: the `IValueFormatter` interface varies based on the target framework, so it must be multiple implementations

Let me know if I can help or if I can make a PR for this issue

StefH:
fix #1084

### DIFF
--- a/src/WireMock.Net.FluentAssertions/Assertions/WireMockAssertions.WithBody.cs
+++ b/src/WireMock.Net.FluentAssertions/Assertions/WireMockAssertions.WithBody.cs
@@ -1,9 +1,12 @@
 #pragma warning disable CS1591
 using System;
 using System.Collections.Generic;
+using AnyOfTypes;
 using Newtonsoft.Json;
 using Newtonsoft.Json.Linq;
+using WireMock.Extensions;
 using WireMock.Matchers;
+using WireMock.Models;
 
 // ReSharper disable once CheckNamespace
 namespace WireMock.FluentAssertions;
@@ -58,7 +61,7 @@ public partial class WireMockAssertions
     {
         var (filter, condition) = BuildFilterAndCondition(r => r.BodyAsBytes, matcher);
 
-        return ExecuteAssertionWithBodyAsBytesExactObjectMatcher(matcher, because, becauseArgs, condition, filter, r => r.BodyAsBytes);
+        return ExecuteAssertionWithBodyAsIObjectMatcher(matcher, because, becauseArgs, condition, filter, r => r.BodyAsBytes);
     }
 
     private AndConstraint<WireMockAssertions> ExecuteAssertionWithBodyStringMatcher(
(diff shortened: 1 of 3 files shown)


src/WireMock.Net.FluentAssertions/Assertions/WireMockAssertions.WithBody.cs
```diff
@@ -148,4 +150,20 @@ public AndConstraint<WireMockAssertions> WithBodyAsBytes(ExactObjectMatcher matc
 
         return new AndConstraint<WireMockAssertions>(this);
     }
+
+    private static string? FormatBody(object? body)
+    {
+        return body switch
+        {
+            null => null,
+            JObject jObject => jObject.ToString(Formatting.None),
```
srollinet, line 159: Since a `JObject` is a `JToken`, and as `ToString()` is not overridden, you could omit this branch.

test/WireMock.Net.Tests/FluentAssertions/WireMockAssertionsTests.cs
```diff
@@ -740,6 +744,41 @@ public async Task HaveReceived1Call_WithBodyAsJson()
...
+            .Should()
+            .HaveReceived(1)
+            .Calls()
+            .WithBodyAsJson(new { x = "y" })
+            .And
+            .UsingPost();
+
+        act.Should()
+            .Throw<Exception>()
+            .WithMessage("Expected wiremockserver to have been called using body \"{\"x\":\"y\"}\", but didn't find it among the body/bodies \"{\"x\":\"123\"}\".");
```
srollinet, line 777: As the `LangVersion` is 11, you could use a raw string literal `"""` to avoid escaping the quotes. It would probably be more readable

srollinet (approved):

StefH:
@srollinet 
Thanks for the review. I'll merge this branch today and create a new NuGet.
And in case you see other things which can be improved, you can create an issue or an PR.

srollinet:
Great, thanks!

PR #1100, opened 2024-05-07
Add support to bind to ip-address instead of only localhost

StefH:


StefH:
https://github.com/WireMock-Net/WireMock.Net/issues/1099

nevaldas:
Sadly, this PR broke usage of standalone WireMock in docker. It spins up, but never answers the requests. Dowgraded one version (before this) - works just fine. We don't specify Urls, just the port.

StefH:
@nevaldas 
Can you specify some more details how you use it and what fails?

Because when I just start the latest WireMock Linux Container, it works fine:
```
docker run -it --rm -p 9091:80 sheyenrath/wiremock.net

Unable to find image 'sheyenrath/wiremock.net:latest' locally
latest: Pulling from sheyenrath/wiremock.net
728328ac3bde: Pull complete
aff7a9c037c8: Pull complete
0a55584cfdc4: Pull complete
36657baceb82: Pull complete
e8b6739a1a24: Pull complete
8344435e42d8: Pull complete
a24fd07a8b7a: Pull complete
Digest: sha256:747f7f26f5a0d60797bb367e828af26b48d9fbead304f961728b5f01d74c87c1
Status: Downloaded newer image for sheyenrath/wiremock.net:latest
05/22/2024 08:21:57 [Info] : By Stef Heyenrath (https://github.com/WireMock-Net/WireMock.Net)
05/22/2024 08:21:57 [Debug] : Server settings {
  "Port": null,
  "UseSSL": null,
  "HostingScheme": null,
  "UseHttp2": false,
  "StartAdminInterface": true,
  "ReadStaticMappings": false,
  "WatchStaticMappings": false,
  "WatchStaticMappingsInSubdirectories": false,
  "ProxyAndRecordSettings": null,
  "Urls": [
    "http://*:80"
  ],
  "StartTimeout": 10000,
  "AllowPartialMapping": false,
  "AdminUsername": null,
  "AdminPassword": null,
  "AdminAzureADTenant": null,
  "AdminAzureADAudience": null,
  "RequestLogExpirationDuration": null,
  "MaxRequestLogCount": null,
  "CorsPolicyOptions": 0,
  "AllowCSharpCodeMatcher": false,
  "AllowBodyForAllHttpMethods": false,
  "AllowOnlyDefinedHttpStatusCodeInResponse": false,
  "DisableJsonBodyParsing": false,
  "DisableRequestBodyDecompressing": false,
  "DisableDeserializeFormUrlEncoded": false,
  "HandleRequestsSynchronously": false,
  "CertificateSettings": null,
  "CustomCertificateDefined": false,
  "ClientCertificateMode": 0,
  "AcceptAnyClientCertificate": false,
  "WebhookSettings": null,
  "UseRegexExtended": true,
  "SaveUnmatchedRequests": false,
  "DoNotSaveDynamicResponseInLogEntry": false,
  "QueryParameterMultipleValueSupport": null,
  "ProtoDefinitions": null,
  "GraphQLSchemas": null
}
05/22/2024 08:21:58 [Info] : Server using .NET 6.0
05/22/2024 08:21:58 Press Ctrl+C to shut down
05/22/2024 08:21:58 WireMock.Net server running
05/22/2024 08:22:28 WireMock.Net server running
05/22/2024 08:22:58 WireMock.Net server running
05/22/2024 08:23:28 WireMock.Net server running
05/22/2024 08:23:58 WireMock.Net server running
```

image

nevaldas:
I am running Docker on Windows. I am forwarding port 8100 to exposed port 80 in the container. 
Container runs just fine, says it's up and ready. Since we specify only port (in this case it's 80), it binds to localhost.
image
Then, calling from my host machine, request fails (but it works with older version).
image
I think, the reason is that kestrel is now listening for localhost only and previous versions used `ListenAnyIp`.

nevaldas:
After digging more into this, we found a difference between our own containers and yours.
As mentioned previously, we were setting only `Port` and that was fine while `ListenAnyIp` was used. 
Now we started using `Urls` which solves our problem, because of wildcard as host (just like in your docker images):
image
Sadly, I still think that this is a breaking change and more similar issues could be reported.

StefH:
My Docker Image provides the URL like this: `ENTRYPOINT ["dotnet", "wiremock-net.dll", "--Urls", "http://*:80"]` for the Linux and Windows version.

You provide only the port?
Like `ENTRYPOINT ["dotnet", "wiremock-net.dll", "--Port", "80"]`

nevaldas:
Our entry point is abit different, but in short - it does not provide port there. It starts our own application which then internally starts wiremock standalone server. And via settings we used to provide only `Port`:
image
Bigger picture:
image

StefH:
When running locally on my PC this works fine:
``` c#
var server = WireMockServer.Start(new WireMockServerSettings
{
    Urls = new[] { "http://*:9091" },
    StartAdminInterface = true
});
System.Console.WriteLine($"3: {string.Join(", ", server.Urls)}");

System.Console.WriteLine("Press any key to stop...");
System.Console.ReadKey();
server.Stop();
```

But in your case, when this same code is running inside a Docker container, it fails?

nevaldas:
Partially yes. If you would change given code to this and run it in Docker container, it would not respond anymore:
```
var server = WireMockServer.Start(new WireMockServerSettings
{
    Port = 9091,
    StartAdminInterface = true
});
System.Console.WriteLine($"3: {string.Join(", ", server.Urls)}");

System.Console.WriteLine("Press any key to stop...");
System.Console.ReadKey();
server.Stop();

StefH:
I can reproduce it.

I think this PR will fix it:
https://github.com/WireMock-Net/WireMock.Net/pull/1107

StefH:
A new NuGet version 1.5.55 will be available in some time.

PR #1132, opened 2024-07-04
Multipart Matcher Fix

rmeshksar:
Fixes the issue discussed here:
https://github.com/WireMock-Net/WireMock.Net/issues/1119

### DIFF
--- a/src/WireMock.Net/Matchers/Request/RequestMessageMultiPartMatcher.cs
+++ b/src/WireMock.Net/Matchers/Request/RequestMessageMultiPartMatcher.cs
@@ -71,23 +71,21 @@ public class RequestMessageMultiPartMatcher : IRequestMatcher
         {
             var mimePartMatchers = Matchers.OfType<MimePartMatcher>().ToArray();
 
-            foreach (var mimePart in message.BodyParts.OfType<MimeKit.MimePart>())
+            foreach (var mimePartMatcher in Matchers.OfType<MimePartMatcher>().ToArray())
             {
-                var matchesForMimePart = new List<MatchResult> { default };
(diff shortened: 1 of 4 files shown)


src/WireMock.Net/Matchers/Request/RequestMessageMultiPartMatcher.cs
```diff
@@ -71,23 +71,21 @@ public double GetMatchingScore(IRequestMessage requestMessage, IRequestMatchResu
...
-                    if (MatchOperator == MatchOperator.Or)
+                    var matchResult = mimePartMatcher.IsMatch(mimeBodyPart);
+                    if (matchResult.IsPerfect())
                     {
+                        score = MatchScores.Perfect;
                         break;
                     }
                 }
-                else
+                if ((MatchOperator == MatchOperator.Or && MatchScores.IsPerfect(score))
```
StefH, line 86: Can be replaced by?
```
if (MatchOperator is MatchOperator.Or or MatchOperator.And && MatchScores.IsPerfect(score))
```

StefH (changes requested):

src/WireMock.Net/Matchers/Request/RequestMessageMultiPartMatcher.cs
StefH: @rmeshksar 
Can you take a look at this comment?

rmeshksar:
@StefH , 
I replied in the conversation for your comment but that message stays in Pending state. 
I repeat it here:


Your suggested if statement and what I did are not the same.
I replicated all possibilities and the two if statement don't match in all cases.

PR #1142, opened 2024-07-22
Make property FromConfiguredStub nullable (for WireMock.Org)

StefH:


crate-scarter:
Thank you for the quick reply.

PR #1175, opened 2024-09-19
Add WireMock.Net.AspNetCore.Middleware

StefH:


src/WireMock.Net.AspNetCore.Middleware/HttpDelegatingHandler/IWireMockDelegationHandlerSettings.cs
```diff
@@ -0,0 +1,8 @@
+// Copyright © WireMock.Net
+
+namespace WireMock.Net.AspNetCore.Middleware.HttpDelegatingHandler;
+
+internal interface IWireMockDelegationHandlerSettings
```
matthewyost, line 5: Why are we creating an interface for this?  This isn't really necessary as how you're registering it in the ServiceCollectionExtensions.cs class could just be done with a standard class (no interface required).  This would reduce the need for 1 extra file.

matthewyost (changes requested):
Just one small change around the need for the interface.  I think we fix that up and it's good to go.

src/WireMock.Net.AspNetCore.Middleware/HttpDelegatingHandler/IWireMockDelegationHandlerSettings.cs
StefH: Good point. I will take a look.

However can you checkout my branch and verify if this works? For me it does not work...

PR #1181, opened 2024-09-25
WireMock.Net.Testcontainers: Use 'sheyenrath/wiremock.net-alpine' image as default for Linux

StefH:


src/WireMock.Net.Testcontainers/WireMockContainerBuilder.cs
```diff
@@ -36,9 +36,11 @@ public sealed class WireMockContainerBuilder : ContainerBuilder<WireMockContaine
         using var dockerClient = dockerClientConfig.CreateClient();
 
         var version = await dockerClient.System.GetVersionAsync();
-        return version.Os.IndexOf("Windows", StringComparison.OrdinalIgnoreCase) > -1;
+        return version.Os.IndexOf("Windows", StringComparison.OrdinalIgnoreCase) > -1 ? OSPlatform.Windows : OSPlatform.Linux;
```
Xor-el, line 39: I see you are checking if the Docker is Windows else you fall back and assume Linux. Is this safe?
wouldn't this better be converted to a switch statement that throws if an Unknown/Unsupported Docker OS is detected, say MacOS?

PR #1200, opened 2024-10-21
Allow adding digital signatures to webhook headers

mathgr22:
Because webhook bodies and headers are transformed separately, Handlebars cannot be used to generate digital signatures using both the body and the header. This change allows the user to specify two callbacks to use to generate a digital signature for the webhook before it is sent.

### DIFF
--- a/src/WireMock.Net.Abstractions/Models/IDigitalSignature.cs
+++ b/src/WireMock.Net.Abstractions/Models/IDigitalSignature.cs
@@ -1,0 +1,41 @@
+// Copyright © WireMock.Net
+
+using System.Threading.Tasks;
+using WireMock.Util;
+
+namespace WireMock.Models;
+
+/// <summary>
+/// A delegate encapsulating logic for fetching or generating a private key.
+/// </summary>
+/// <returns>An async task containing the value of the private key.</returns>
+public delegate Task<string> PrivateKeyCallback();
+
+/// <summary>
+/// A delegate encapsulating logic for fetching or generating a digital signature.
+/// </summary>
+/// <param name="privateKey">The value of the private key returned by <see cref="PrivateKeyCallback"/>.</param>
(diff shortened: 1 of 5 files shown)


StefH:
@mathgr22 
Thanks for this PR, however this is very specific to your use-case.

I think it's better to add functionality to support providing your own callback method to generate the complete request.

A bit look-a-like as this:
https://github.com/WireMock-Net/WireMock.Net/blob/master/src/WireMock.Net/ResponseBuilders/Response.WithBody.cs

mathgr22:
@StefH
You're suggesting a WebhookRequest.WithBody, or something to that effect? This was my original idea, unfortunately my tech-lead thought something specific to digital signatures was a better route. I could certainly attempt to implement something along the lines of what you suggested.

StefH:
Instead of WebhookRequest.WithBody, I think this should be WebhookRequest.WithXXX (I don't know a good name now)
But essentially the parameter should be a Func/Action which takes the request and maybe other fields and from that data, you should create a complete response in C# code (complete = body + headers + all else)

mathgr22:
I think the problem I'm running into here is that I want to be able to modify the headers _after_ the body has already been transformed. After implementing a method taking an `IRequestMessage` as a parameter, I realized that this doesn't do anything for me, as I need the body to already have been transformed before I can generate the digital signature. Of course I could not use transformers, but it would be nice to be able to use transformers as well.

Essentially, I need something like `IWebhookRequest.WithPostTransformProcess(Func<IWebhookRequest, IWebhookRequest> webhookPostProcesser)` that will allow me to intercept the webhook after it's transformed, but before it's sent. This could be done through a more generalized version of the code in this PR, but I'm not too sure what the signatures should look like, etc.

mathgr22:
Or perhaps I need an event on the `IWebhookRequest` that runs before sending `.OnBeforeSending` or something.

PR #1248, opened 2025-01-29
Add exception message to logging when mapping fails due to an exception.

JvE-iO:
I spent way too much time in the debugger trying to figure out why the mapping was not found. 
It was found, but an exception occurred during the mapping, leading to a misleading message.

As an aside, the exception occurred because I called `.WithMapping()` before I called `.AddProtoDefinition()`...

### DIFF
--- a/src/WireMock.Net/Owin/WireMockMiddleware.cs
+++ b/src/WireMock.Net/Owin/WireMockMiddleware.cs
@@ -224,7 +224,7 @@ namespace WireMock.Owin
                 }
                 catch (Exception ex)
                 {
-                    _options.Logger.Error("HttpStatusCode set to 404 : No matching mapping found", ex);
+                    _options.Logger.Error("HttpStatusCode set to 404 : 'No matching mapping found', due to exception '{0}'", ex);
 
                     var notFoundResponse = ResponseMessageBuilder.Create(HttpStatusCode.NotFound, WireMockConstants.NoMatchingFound);
                     await _responseMapper.MapAsync(notFoundResponse, ctx.Response).ConfigureAwait(false);
--- a/src/WireMock.Net/Owin/WireMockMiddleware.cs
+++ b/src/WireMock.Net/Owin/WireMockMiddleware.cs
@@ -224,7 +224,7 @@ namespace WireMock.Owin
                 }
                 catch (Exception ex)
                 {
-                    _options.Logger.Error("HttpStatusCode set to 404 : 'No matching mapping found', due to exception '{0}'", ex);
+                    _options.Logger.Error("HttpStatusCode set to 404 : No matching mapping found", ex);
 
                     var notFoundResponse = ResponseMessageBuilder.Create(HttpStatusCode.NotFound, WireMockConstants.NoMatchingFound);
                     await _responseMapper.MapAsync(notFoundResponse, ctx.Response).ConfigureAwait(false);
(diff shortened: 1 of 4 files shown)


src/WireMock.Net/Owin/WireMockMiddleware.cs
```diff
@@ -224,7 +224,7 @@ private async Task InvokeInternalAsync(IContext ctx)
                 }
                 catch (Exception ex)
                 {
-                    _options.Logger.Error("HttpStatusCode set to 404 : No matching mapping found", ex);
+                    _options.Logger.Error("HttpStatusCode set to 404 : 'No matching mapping found', due to exception '{0}'", ex);
```
StefH, line 227: There is also Error method with an exception. Would that also work for you?

https://github.com/WireMock-Net/WireMock.Net/blob/master/src/WireMock.Net.Abstractions/Logging/IWireMockLogger.cs#L57

StefH (changes requested):

src/WireMock.Net/Owin/WireMockMiddleware.cs
JvE-iO: I think that's already being used.

But `WireMock.Net.Xunit.TestOutputHelperWireMockLogger.Error(string formatString, Exception exception)` expects the formatString to have a placeholder for the exception message.

So it could also be fixed there. And in `WireMock.Logging.WireMockConsoleLogger.Error(string formatString, Exception exception)`.

StefH: OK I see, I did not detect this...

---

However, the reason that I did not expose the exception is maybe because of security.

The best way to detect if a mapping is working correct or incorrect is looking at the logging.
https://github.com/WireMock-Net/WireMock.Net/wiki/Request-Matching-Tips

JvE-iO: Yes, that eventually told me it was a perfect match.

But that was immediately followed by a "HttpStatusCode set to 404 : No matching mapping found".

And then you're in limbo.
WireMock says it has a matching mapping and WireMock says no matching mapping was found. Nothing else is logged.

I finally figured out what was going on by creating and injecting a custom IWireMockLogger that does log exception messages.

Not logging the exception message for security reasons sounds strange to me. How would anyone know that an exception has occurred then? Also, your loggers *do* log all the exception messages when the type is AggregateException.

Anyway, even changing it to something like
```csharp
_options.Logger.Error("Mapping failed due to unhandled exception. HttpStatusCode set to 404 : No matching mapping found", ex);
```
may help someone in the future a little bit.

Or maybe an exception occuring after a matching mapping has been found should not return a 404 with a "No matching mapping found" message. A 500 "something went wrong during the mapping" is probably more helpful.

StefH: Maybe I miss the context, but this line **does** log the exception?
`_options.Logger.Error("HttpStatusCode set to 404 : No matching mapping found", ex);`

Example:
```c#
Error("HttpStatusCode set to 404 : No matching mapping found", new ArgumentNullException("abc"));

return;

void Error(string message, Exception exception)
{
    Console.WriteLine(Format("Error", $"{message} {{0}}", exception));

    if (exception is AggregateException ae)
    {
        ae.Handle(ex =>
        {
            Console.WriteLine(Format("Error", "Exception {0}", ex));
            return true;
        });
    }
}

static string Format(string level, string formatString, params object[] args)
{
    var message = args.Length > 0 ? string.Format(formatString, args) : formatString;

    return $"{DateTime.UtcNow} [{level}] : {message}";
}
```

Logging is:
```
1/29/2025 3:50:30 PM [Error] : HttpStatusCode set to 404 : No matching mapping found System.ArgumentNullException: Value cannot be null. (Parameter 'abc')
```

StefH: But the TestOutputHelperWireMockLogger.cs could have an issue ?

StefH (approved):

PR #1263, opened 2025-03-13
Add ProtoDefinitionHelper.FromDirectory

StefH:


src/WireMock.Net/Util/ProtoDefinitionHelper.cs
```diff
@@ -1,12 +1,50 @@
...
+
+            // Build comment and get content from file.
+            var comment = $"// {relativePath}";
+            var content = File.ReadAllText(filePath);
+
+            // Only add the comment if it's not already defined.
+            var modifiedContent = !content.StartsWith(comment) ? $"{comment}\n{content}" : content;
+            var key = Path.GetFileNameWithoutExtension(filePath);
+
+            dictionary.Add(key, modifiedContent);
```
siewers, line 42: I'm curious why the filename isn't used as the key here. That would make resolving imports work without the comment.
The only change needed is to replace `\` with `/` of the relative path.

StefH: > I'm curious why the filename isn't used as the key here.
The filename is used: `var key = Path.GetFileNameWithoutExtension(filePath);`

> That would make resolving imports work without the comment.
The IProtoFileResolver is used in [ProtobufJsonConverterFileSystem.cs](https://github.com/StefH/ProtoBufJsonConverter/blob/main/src/ProtoBufJsonConverter/IO/ProtobufJsonConverterFileSystem.cs) which is required by protobuf-net to resolve additional files.

siewers: From my debugging, it tries to resolve based on the import filename, which would match the key in the dictionary based on the relative file path, no?

Issue #1243, opened 2025-01-22
Support for AwesomeAssertions


jm-atbas:
[FluentAssertions](https://github.com/fluentassertions/fluentassertions) recently switched to a commercial license: https://github.com/fluentassertions/fluentassertions/pull/2943 starting from version `8.X`. This has led many looking for an alternative. One possible alternative has now emerged in the form of a community fork called [AwesomeAssertions](https://github.com/AwesomeAssertions/AwesomeAssertions).

I would like WireMock to offer a NuGet package `WireMock.Net.AwesomeAssertions` to support this new library. It would essentially be a copy of the already existing integration for FluentAssertions since they are fully compatible. I would be open to create PR with the necessary changes.

jcfnomada:
I do like this idea. It's actually just a copy of the same project, and changing the package from FluentAssertions to AwesomeAssertions (8.0.0). The namespaces are the same.

PR #1273, opened 2025-04-09
feat(awesome-assertions): Added new project WireMock.Net.AwesomeAssertions

Crowbar90:
Adding new project for AwesomeAssertions support after FluentAssertions license change (as proposed in #1243).

### DIFF
--- a/src/WireMock.Net.AwesomeAssertions/Assertions/WireMockANumberOfCallsAssertions.cs
+++ b/src/WireMock.Net.AwesomeAssertions/Assertions/WireMockANumberOfCallsAssertions.cs
@@ -1,0 +1,40 @@
+// Copyright © WireMock.Net
+
+using Stef.Validation;
+using WireMock.Server;
+
+// ReSharper disable once CheckNamespace
+namespace WireMock.FluentAssertions;
+
+/// <summary>
+/// Provides assertion methods to verify the number of calls made to a WireMock server.
+/// This class is used in the context of FluentAssertions.
+/// </summary>
+public class WireMockANumberOfCallsAssertions
+{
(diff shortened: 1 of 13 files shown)


src/WireMock.Net.AwesomeAssertions/Assertions/WireMockANumberOfCallsAssertions.cs
StefH, line 1: It think this file is the same as in the other project.

Can you just link it instead of making a copy?

Same applies to all .cs files

StefH (changes requested):

src/WireMock.Net.AwesomeAssertions/Assertions/WireMockANumberOfCallsAssertions.cs
Crowbar90: Actually there is a small difference, `Execute.Assertion` was replaced with the passing of a `AssertionChain` object in AwesomeAssertions (see https://awesomeassertions.org/upgradingtov8#from-executeassertion-to-assertionchain).

However I will check and link the files which don't have changes.

Crowbar90: After checking, all files have been affected by this change. Anyway I pushed another commit in order to adhere to the dotnet naming convention for private fields.

StefH: I see.

However, the code is now duplicated, so in case of bug-fix or new feature, this needs to be updated in 2 places

StefH (approved):

Issue #1284, opened 2025-04-24
WireMock.Net.AspNetCore.Middleware does not work in hosted services


etkr:
### Describe the bug

WireMock AspNetCore Middleware works great when a request is sent from within a request thread, but if a request is sent from a background service it throws this exception: 

```
System.ArgumentNullException: Value cannot be null. (Parameter '_httpContextAccessor.HttpContext')
   at Stef.Validation.Guard.NotNull[T](T value, String parameterName)
   at WireMock.Net.AspNetCore.Middleware.HttpDelegatingHandler.WireMockDelegationHandler.SendAsync(HttpRequestMessage request, CancellationToken cancellationToken)
   at Microsoft.Extensions.Http.Logging.LoggingScopeHttpMessageHandler.<SendCoreAsync>g__Core|5_0(HttpRequestMessage request, Boolean useAsync, CancellationToken cancellationToken)
   at System.Net.Http.HttpClient.<SendAsync>g__Core|83_0(HttpRequestMessage request, HttpCompletionOption completionOption, CancellationTokenSource cts, Boolean disposeCts, CancellationTokenSource pendingRequestsCts, CancellationToken originalCancellationToken)
```

What is happening is that `HttpContext` is not available in a hosted service, it is a nullable property of `IHttpContextAccessor` since it is not always guaranteed to be available.

### Expected behavior:

It should be possible to make http requests in a background service while WireMock Middleware is enabled.

Additionally it should be possbile to mock requests coming from a background service.

### Test to reproduce
* Enable WireMock AspNetCore Middleware with `AddWireMockService`
* Make any http request in a hosted service
* Oberve that `ArgumentNullException` is thrown from `HttpClient`

### Other related info

I created a PR #1285 with a test that reproduces this bug and a proposed solution.

PR #1285, opened 2025-04-24
Enable support for WireMock Middleware in Hosted Services

etkr:
## Background

While attempting to use Wiremock Middle in an application that uses hosted services i noticed that it breaks and throws an `ArgumentNullException` .

See #1284 

# Problem 

The problem is that `IHttpContextAccessor.HttpContext` can not be relied upon to always be available.

It would be best if `WireMockDelegationHandler` can fallback gracefully if `HttpContext` is null, which is expected in some cases.

# Solution

I changed `IsWireMockRedirectHeaderSetToTrue` and `TryGetDelayHeaderValue` in `WireMockDelegationHandler` to default to false if `HttpContext` is null.

I created a test to check that it works when started in a hosted service.

# Limitations

The `X-WireMock-Redirect` and `X-WireMock-Response-Delay` headers are obviously not going to work as expected when mocking a request from a background service, maybe this needs to be documented somewhere.

### DIFF
--- a/src/WireMock.Net.AspNetCore.Middleware/HttpDelegatingHandler/WireMockDelegationHandler.cs
+++ b/src/WireMock.Net.AspNetCore.Middleware/HttpDelegatingHandler/WireMockDelegationHandler.cs
@@ -36,7 +36,6 @@ internal class WireMockDelegationHandler : DelegatingHandler
     protected override async Task<HttpResponseMessage> SendAsync(HttpRequestMessage request, CancellationToken cancellationToken)
     {
         Guard.NotNull(request);
-        Guard.NotNull(_httpContextAccessor.HttpContext);
 
         if (_settings.AlwaysRedirect || IsWireMockRedirectHeaderSetToTrue())
         {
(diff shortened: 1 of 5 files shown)


src/WireMock.Net.AspNetCore.Middleware/HttpDelegatingHandler/WireMockDelegationHandler.cs
```diff
@@ -57,16 +56,20 @@ protected override async Task<HttpResponseMessage> SendAsync(HttpRequestMessage
 
     private bool IsWireMockRedirectHeaderSetToTrue()
     {
+        var httpContext = _httpContextAccessor.HttpContext;
+        if (httpContext is null) return false;
```
StefH, line 60: Can you please use `{` and `}` for if-statements?

And is it possible to add logging here in case the httpContext == null ?

StefH (changes requested):

src/WireMock.Net.AspNetCore.Middleware/HttpDelegatingHandler/WireMockDelegationHandler.cs
etkr: Add braces and debug logging

StefH (approved):

PR #1301, opened 2025-05-21
Fix errors when converting response bodies with top-level JSON arrays to C# code

asaf92:
When trying to convert static mapping to C# code, I ran into an exception when the body is a top-level array rather than a pure string or a (non-array) JSON object.

The problem is that this code doesn't distinguish between arrays and objects, and assumes that every non-null that isn't a string is an object:
```cs
case BodyType.Json:
    if (bodyData.BodyAsJson is string bodyStringValue)
    {
        sb.AppendLine($"        .WithBody({ToCSharpStringLiteral(bodyStringValue)})");
    }
    else if (bodyData.BodyAsJson is { } jsonBody)
    {
        var anonymousObjectDefinition = ConvertToAnonymousObjectDefinition(jsonBody);
        sb.AppendLine($"        .WithBodyAsJson({anonymousObjectDefinition})");
    }
                                                                                      
    break;
```

My solution is to load a JToken instead of a JObject in ConvertToAnonymousObjectDefinition, which seems like the original intention given that ConvertJsonToAnonymousObjectDefinition (which is defined in an internal class) expects a JToken and isn't called anywhere else in the code.

### DIFF
--- a/src/WireMock.Net/Util/CSharpFormatter.cs
+++ b/src/WireMock.Net/Util/CSharpFormatter.cs
@@ -96,13 +96,13 @@ internal static class CSharpFormatter
         "while"
     });
     #endregion
-    
+
     public static object ConvertToAnonymousObjectDefinition(object jsonBody, int ind = 2)
     {
         var serializedBody = JsonConvert.SerializeObject(jsonBody);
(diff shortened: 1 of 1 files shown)


StefH (changes requested):
Can you also add a unit test?

StefH (approved):

Issue #1321, opened 2025-06-18
Feature: Setting individual scenario state via Admin API


noxfate:
**Problem**
I'm currently using static file mapping as the main source for development, integration-test and e2e testing with the same set of static file mapping. It has grow in size and number of files has become more difficult to manage, but the pros of it still appreciated. This have cause flakiness and harder to do mock in isolation. By forcing scenario state really help solve this problem.

I used to leverage stateful behaviour and manual setting individual state, not only for stability but for test data maintainability and make sure that the mock is in isolation as well. I do think this really help with mock data traceability and maintainability of the test.

![Image](https://github.com/user-attachments/assets/a55fe409-104c-4d00-8876-2852814c8ace)

**Describe the solution you'd like**
Java implementation of wiremock has this endpoint POST /__admin/scenarios/<scenario_name>/state available in admin api client. I was hoping this become available in this .Net version as well. 
ref: https://wiremock.org/docs/stateful-behaviour/#setting-the-state-of-an-individual-scenario

**Describe alternatives you've considered**
We have tried to organise the static file by using "priority" but that become a nightmare to handle real quick.

**Is your feature request supported by [WireMock (java version)](https://www.wiremock.org)? Please provide details.**
Yes. here is the clear doc https://wiremock.org/docs/stateful-behaviour/#setting-the-state-of-an-individual-scenario

**Additional context**
N/A

StefH:
@noxfate 
I think I've updated the code to support this.
Can you try preview version: WireMock.1.8.12-ci-110218

The API call is a PUT with body:
``` json
{
  "State": "the state"
}
```


(https://github.com/wiremock/WireMock.Net/wiki/MyGet-preview-versions)

noxfate:
Thank you very much for a fast support! A couple notes

I cannot make it to work yet, seems like a problem not on the WireMock.Net implementation anymore but on the docker image that I was using. Im mainly testing on WireMock.Net.TestContainers package and the server that I tried to curl directly to is not that up-to-date. Any suggestions where I can take the correlate update of the image?

I tested with the image tag in the screenshot above `sheyenrath/wiremock.net-alpine:1.8.10`. This is the result from my debug directly postman to the testcontainers.
![Image](https://github.com/user-attachments/assets/96eadb99-a933-4b89-a8f6-470e8f20181c)

Meanwhile, I'll try to setup quickly on the WireMock standalone to help test your implementation.

Few thing that I hope its useful
[...]

PR #1322, opened 2025-06-19
Add Scenario set State method

StefH:


StefH:
https://github.com/wiremock/WireMock.Net/issues/1321

PR #1331, opened 2025-07-16
Correctly map the Pact Interaction Description property

StefH:


basdijkstra:
@StefH yup, this is what I was looking for. I'd still have to test it properly when you have released this fix, but it looks good.

I wanted to publish the pact that you generated, the `PactTests.SavePact_Get_Request_And_Response_WithBodyAsJson.verified.json` one, but my Pact Broker complained it wasn't valid. No idea why, jsonlint.com thought it was fine. Changing `description` back to `providerState` wasn't the solution so it must be something else.

Again, this looks good, I'm happy to test it with my own contract and code once you've released the update.

StefH:
I would think that the generated JSON file would be valid? Can you more details on why it fails?

basdijkstra:
This is the command I use to publish the contract to the broker:

`pact-broker publish PactTests.SavePact_Get_Request_And_Response_WithBodyAsJson.verified.json --consumer-app-version 1.0.0`

and this is the output from the `pact-broker` command:
```
C:/Tools/pact/lib/ruby/lib/ruby/3.3.0/json/common.rb:219:in `parse': unexpected token at '﻿{ (JSON::ParserError)
  "consumer": {
    "name": "Something API Consumer Get"
  },
  "interactions": [
    {
      "description": "A GET request to retrieve the something",
      "request": {
        "headers": {
          "Accept": "application/json"
        },
        "method": "GET",
        "path": "/tester",
        "query": "q1=test&q2=ok"
      },
      "response": {
        "body": {
          "id": "tester",
          "firstName": "Totally",
          "lastName": "Awesome"
        },
        "headers": {
          "Content-Type": "application/json; charset=utf-8"
        },
        "status": 200
      }
    }
  ],
  "provider": {
    "name": "Something API"
  }
}'
        from C:/Tools/pact/lib/ruby/lib/ruby/3.3.0/json/common.rb:219:in `parse'
        from C:/Tools/pact/lib/vendor/ruby/3.3.0/gems/pact_broker-client-1.77.0/lib/pact_broker/client/pact_file.rb:28:in `pact_hash'
        from C:/Tools/pact/lib/vendor/ruby/3.3.0/gems/pact_broker-client-1.77.0/lib/pact_broker/client/pact_file.rb:20:in `consumer_name'
...
```
I really can't see anything that's wrong with the .json file.

I also tried publishing another contract, generated with an earlier version of WireMock .Net, and manually replace `providerState` with `description`, and that led to the desired result, so I think the code change itself is OK:

image

StefH:
If you want to test wiremock, you can try preview version 1.8.14-ci-110308

See
https://github.com/wiremock/WireMock.Net/wiki/MyGet-preview-versions

basdijkstra:
I'll set a reminder and do that tomorrow morning!

basdijkstra:
@StefH I used this new preview version for a test with my code and it worked like a charm:
image

Thanks so much, ready to release this as far as I'm concerned.

Issue #1338, opened 2025-07-27
Specifying .WithBody(byte[]) fails to match for JSON bodies


smfields:
### Describe the bug
When using the overload of `.WithBody()` that accepts a `byte[]` it will fail to match on requests that have a matching body if the bytes happen to correspond to valid JSON.

The behavior is inconsistent as it will correctly match on arbitrary strings that have been encoded and provided as bytes, but it will fail to match if the string corresponds to valid JSON, regardless of the content type specified on the request.

### Expected behavior:
WireMock should match requests that have a body that matches the provided bytes, regardless of what those bytes represent.

### Test to reproduce
See repro: https://github.com/smfields/WiremockIssueRepro/tree/master

### Other related info
Seems like the issue probably stems from [this section](https://github.com/wiremock/WireMock.Net/blob/e400e92452da3d1e46c3d97d24aca56dd5ad89b0/src/WireMock.Net.Shared/Matchers/Helpers/BodyDataMatchScoreCalculator.cs#L37) of `BodyDataMatchScoreCalculator.cs`.

When the matcher is an `ExactObjectMatcher` it will compare the value against the bytes only if the detected body type is bytes, string, or form. When the detected body type is JSON we drop down to the next section that compares the provided value against the `BodyAsJson`, but since the provided value is bytes this doesn't line up. 

Seems like we need to either open up the comparison on line 41 to also include times when the provided value is of type `byte[]`, or we need to parse the bytes into JSON before doing the comparison against `BodyAsJson`.

smfields:
Might be as easy as what I've started in https://github.com/wiremock/WireMock.Net/pull/1339, but more testing would be needed to confirm.

PR #1339, opened 2025-07-27
Fixes an issue with matching JSON bodies as bytes

smfields:
Fixes an issue where body payloads would fail to be matched if the matcher was created using a `byte[]` value and the bytes represented valid JSON.

## References

- #1338 

## Submitter checklist

- [ ] Test coverage that demonstrates that the change works as expected

### DIFF
--- a/src/WireMock.Net.Shared/Matchers/Helpers/BodyDataMatchScoreCalculator.cs
+++ b/src/WireMock.Net.Shared/Matchers/Helpers/BodyDataMatchScoreCalculator.cs
@@ -34,14 +34,10 @@ internal static class BodyDataMatchScoreCalculator
             }
         }
 
-        if (matcher is ExactObjectMatcher exactObjectMatcher)
+        if (matcher is ExactObjectMatcher { Value: byte[] } exactObjectMatcher)
         {
             // If the body is a byte array, try to match.
-            var detectedBodyType = requestMessage.DetectedBodyType;
-            if (detectedBodyType is BodyType.Bytes or BodyType.String or BodyType.FormUrlEncoded || exactObjectMatcher.Value is byte[])
-            {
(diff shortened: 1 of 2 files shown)


StefH (changes requested):
Can you add a new unit test for BodyDataMatchScoreCalculator which tests this scenario?

smfields:
@StefH I added a few new tests.

`RequestMessageBodyMatcher_GetMatchingScore_BodyTypeJson_BodyAsBytes_ExactObjectMapper` is the one that directly covers the scenario this PR addresses.

The other two are for covering the existing scenario of matching non-JSON strings or bytes.

src/WireMock.Net.Shared/Matchers/Helpers/BodyDataMatchScoreCalculator.cs
```diff
@@ -38,7 +38,7 @@ internal static MatchResult CalculateMatchScore(IBodyData? requestMessage, IMatc
         {
             // If the body is a byte array, try to match.
             var detectedBodyType = requestMessage.DetectedBodyType;
-            if (detectedBodyType is BodyType.Bytes or BodyType.String or BodyType.FormUrlEncoded)
+            if (detectedBodyType is BodyType.Bytes or BodyType.String or BodyType.FormUrlEncoded || exactObjectMatcher.Value is byte[])
```
StefH: I'm wondering if this can be replaced by a simpler one-liner? (and skip all logic related to detected body type).
```
if (exactObjectMatcher.Value is byte[])
```

sfields-metabolon: This has been updated

StefH (approved):

PR #1359, opened 2025-08-31
Version 2.x

StefH:
Version 2.x supports:
- .NET Framework 4.6.2
- .NET Standard 2.0 / 2.1
- .NET 8.0

### DIFF
--- a/src/WireMock.Net.Minimal/WebSockets/WebSocketConnectionRegistry.cs
+++ b/src/WireMock.Net.Minimal/WebSockets/WebSocketConnectionRegistry.cs
@@ -2,7 +2,6 @@
 
 using System.Collections.Concurrent;
 using System.Diagnostics.CodeAnalysis;
-using System.Linq;
 using System.Net.WebSockets;
 
 namespace WireMock.WebSockets;
--- a/src/WireMock.Net.Minimal/WebSockets/WebSocketMessageConditionBuilder.cs
+++ b/src/WireMock.Net.Minimal/WebSockets/WebSocketMessageConditionBuilder.cs
@@ -5,24 +5,15 @@ using Stef.Validation;
 
 namespace WireMock.WebSockets;
 
-internal class WebSocketMessageConditionBuilder : IWebSocketMessageConditionBuilder
+internal class WebSocketMessageConditionBuilder(WebSocketBuilder parent, IMatcher matcher) : IWebSocketMessageConditionBuilder
 {
-    private readonly WebSocketBuilder _parent;
-    private readonly IMatcher _matcher;
-
-    public WebSocketMessageConditionBuilder(WebSocketBuilder parent, IMatcher matcher)
-    {
-        _parent = Guard.NotNull(parent);
-        _matcher = Guard.NotNull(matcher);
-    }
-
     public IWebSocketBuilder ThenSendMessage(Action<IWebSocketMessageBuilder> configure)
     {
         Guard.NotNull(configure);
         var messageBuilder = new WebSocketMessageBuilder();
         configure(messageBuilder);
 
-        return _parent.AddConditionalMessage(_matcher, messageBuilder);
+        return parent.AddConditionalMessage(matcher, messageBuilder);
     }
 
     public IWebSocketBuilder SendMessages(Action<IWebSocketMessagesBuilder> configure)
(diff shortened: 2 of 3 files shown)


Kielek:
@StefH, some feedback:
1. It will be great to still have support for .NET Framework 4.6.2. It has [normal support until Jan 12, 2027](https://learn.microsoft.com/en-us/lifecycle/products/microsoft-net-framework). The downside - there is no possibility to update to XUnit.v3. This package is used under [OpenTelemetry Contrib](https://github.com/open-telemetry/opentelemetry-dotnet-contrib) repository. And various vendors still needs to offer support for ancient frameworks.
2. I have executed build locally, there is a lot of NU1902, NU1903, NU1904 warnings. All of them are related to different levels o security vulnerabilities. Part of them affects _production_ dependencies.

StefH:
Thanks for your response.

About 1: WireMock.Net is mostly used in unit tests. So how many % of those unit tests project are going to use an older framework like the one you mentioned? I think that supporting older frameworks in a main project and testing this in a unit test project is related but not always 1 to 1....

I will think on this and maybe support that one instead of 4.8


About 2: the question is how critical any security issue is because WireMock is used for testing, not for production code.

Kielek:
> Thanks for your response.
> 
> About 1: WireMock.Net is mostly used in unit tests. So how many % of those unit tests project are going to use an older framework like the one you mentioned? I think that supporting older frameworks in a main project and testing this in a unit test project is related but not always 1 to 1....
> 
> I will think on this and maybe support that one instead of 4.8

It will be great. I have had similar discussions about other, testing only library - Verify.  After 1,5 year author bringed back support for this old framework. I think that the requestor was .NET team.

> About 2: the question is how critical any security issue is because WireMock is used for testing, not for production code.

I think that here are 2 aspects to consider:

.NET SDK 9 for first release switchted settings to (then it was rollbacked, but some project kept it in place). I am not sure what is the current state:
```xml
  <PropertyGroup>
    <NuGetAudit>true</NuGetAudit>
    <NuGetAuditMode>all</NuGetAuditMode>
    <NuGetAuditLevel>low</NuGetAuditLevel>
  </PropertyGroup>
```
With this + treat warnings as errors, which is pretty common, at least in my buble, it leads to the breaking compilation. Usually, there is a possibility to fix such changes basicaly by bumping transitive packages, but it is just inconvinient. IMO it is better to fix such cases centrally,

The second aspect are security scans. I will consider OpenTelemetry as an example. Some end-users/companies making additional source code scans, not only shipped libraries. If we have any known vulnerable dependencies, there are at least yellow flags and we need to explain that the production code is not affected. With this, this project has strict rules to fix all such warnings.

Thanks for maintaining this library, I know that it takes more time than people usually thinks.

StefH:
@Kielek 
I checked your idead about 4.6.2, however that brings back again some if-defs and other classes which I would like to get rid of. So I'll stick with Framework 4.8 and .NET 8

Kielek:
That's fine, I think that I will need to create some simplified version of this package tailored to our usecase.

StefH:
@petrroll 
This V2 branch only support .NET 48 and .NET 8 , so the OpenTelemetry integration is much easier. Can you take a look at this PR for the OpenTelemetry code?

petrroll:
@StefH I have briefly tried to take a look at it seems you haven't changed many places, just collapsed two settings into one which is great/fine. 

It is however quite a humongous PR so I can't promise I didn't miss anything 😅

Should be fairly easy to test via the test app tho / with producing test docker container and then trying the aspire integration

petrroll:
Both codex 52 and Claude Opus 4.5 agree (thank you my employer for those tokens), so I think we're golden (I know I know, it's not great practice, but skimming it myself + validating with two separate models gives some signal)

StefH:
@Kielek 

I've changed the project ; it will now support

- .NET Framework 4.6.2
- .NET Standard 2.0 / 2.1
- .NET 8.0

Kielek:
> I've changed the project ; it will now support
> 
> * .NET Framework 4.6.2
> * .NET Standard 2.0 / 2.1
> * .NET 8.0

Thanks. I think that it is good decision, but OTel-contrib already dropped this dependency: https://github.com/open-telemetry/opentelemetry-dotnet-contrib/pull/3645

StefH:
> > I've changed the project ; it will now support
> > 
> > * .NET Framework 4.6.2
> > * .NET Standard 2.0 / 2.1
> > * .NET 8.0
> 
> Thanks. I think that it is good decision, but OTel-contrib already dropped this dependency: [open-telemetry/opentelemetry-dotnet-contrib#3645](https://github.com/open-telemetry/opentelemetry-dotnet-contrib/pull/3645)

My OpenTelemetry project will have netstandard2.0 and net8.0

PR #1375, opened 2025-11-05
Add WireMockHealthCheck in WireMock.Net.Aspire

Zguy:
I know the contributing guidelines says to make an issue first to have a discussion, but we can have the discussion here and I don't mind if the effort was wasted.

The point of this change is to make `WaitFor()` in Aspire more useful with WireMock, since it currently doesn't wait for anything. This can be a problem and create a race condition if the main project starts doing HTTP calls to the wiremock server immediately on startup, and will on some runs get a 404 back.

To fix that I've added a health check that calls `/__admin/health` and checks the result, as well as checks if mappings using AdminApiMappingBuilder has been submitted to the server.

That last part created a catch-22 problem where the mappings were not submitted until the health check was healthy, but the health check was not healthy until the mappings were submitted.

To avoid this, the WireMockServerLifecycleHook class has been slightly re-arranged, and is now using the AfterEndpointsAllocatedAsync callback rather than the AfterResourcesCreatedAsync callback. Within which a separate Task is created that waits until the server is ready and submits the mappings.

## Submitter checklist

### DIFF
--- a/src/WireMock.Net.Aspire/WireMockServerLifecycleHook.cs
+++ b/src/WireMock.Net.Aspire/WireMockServerLifecycleHook.cs
@@ -10,11 +10,12 @@ internal class WireMockServerLifecycleHook(ILoggerFactory loggerFactory) : IDist
 {
     private readonly CancellationTokenSource _shutdownCts = new();
 
+    private CancellationTokenSource? _linkedCts;
     private Task? _mappingTask;
 
-    public async Task AfterEndpointsAllocatedAsync(DistributedApplicationModel appModel, CancellationToken cancellationToken = default)
+    public Task AfterEndpointsAllocatedAsync(DistributedApplicationModel appModel, CancellationToken cancellationToken = default)
     {
-        var cts = CancellationTokenSource.CreateLinkedTokenSource(_shutdownCts.Token, cancellationToken);
+        _linkedCts = CancellationTokenSource.CreateLinkedTokenSource(_shutdownCts.Token, cancellationToken);
 
         _mappingTask = Task.Run(async () =>
         {
@@ -29,19 +30,27 @@ internal class WireMockServerLifecycleHook(ILoggerFactory loggerFactory) : IDist
                 var endpoint = wireMockServerResource.GetEndpoint();
                 System.Diagnostics.Debug.Assert(endpoint.IsAllocated);
 
-                await wireMockServerResource.WaitForHealthAsync(cts.Token);
+                await wireMockServerResource.WaitForHealthAsync(_linkedCts.Token);
 
-                await wireMockServerResource.CallApiMappingBuilderActionAsync(cts.Token);
+                await wireMockServerResource.CallApiMappingBuilderActionAsync(_linkedCts.Token);
 
-                wireMockServerResource.StartWatchingStaticMappings(cts.Token);
+                wireMockServerResource.StartWatchingStaticMappings(_linkedCts.Token);
             }
-        }, cts.Token);
+        }, _linkedCts.Token);
+
+        return Task.CompletedTask;
     }
 
     public async ValueTask DisposeAsync()
     {
         await _shutdownCts.CancelAsync();
+
+        _linkedCts?.Dispose();
+        _shutdownCts.Dispose();
+
         if (_mappingTask is not null)
+        {
             await _mappingTask;
+        }
     }
 }


src/WireMock.Net.Aspire/WireMockServerLifecycleHook.cs
```diff
@@ -10,32 +10,38 @@ internal class WireMockServerLifecycleHook(ILoggerFactory loggerFactory) : IDist
...
                 wireMockServerResource.StartWatchingStaticMappings(cts.Token);
             }
-        }
+        }, cts.Token);
     }
 
     public async ValueTask DisposeAsync()
     {
         await _shutdownCts.CancelAsync();
+        if (_mappingTask is not null)
```
StefH, line 51: can you add { and { for if statements?

src/WireMock.Net.Aspire/WireMockServerResource.cs
```diff
@@ -20,6 +20,14 @@ public class WireMockServerResource : ContainerResource, IResourceWithServiceDis
     internal WireMockServerArguments Arguments { get; }
     internal Lazy<IWireMockAdminApi> AdminApi => new(CreateWireMockAdminApi);
 
+    internal enum MappingState
```
StefH: Can you move this enum to its own file?

src/WireMock.Net.Aspire/WireMockServerLifecycleHook.cs
```diff
@@ -10,32 +10,38 @@ internal class WireMockServerLifecycleHook(ILoggerFactory loggerFactory) : IDist
 {
     private readonly CancellationTokenSource _shutdownCts = new();
 
-    public async Task AfterResourcesCreatedAsync(DistributedApplicationModel appModel, CancellationToken cancellationToken = default)
+    private Task? _mappingTask;
+
+    public async Task AfterEndpointsAllocatedAsync(DistributedApplicationModel appModel, CancellationToken cancellationToken = default)
     {
         var cts = CancellationTokenSource.CreateLinkedTokenSource(_shutdownCts.Token, cancellationToken);
```
StefH: Sonar complains on this.

Maybe dispose this cts in DisposeAsync method?

src/WireMock.Net.Aspire/WireMockServerLifecycleHook.cs
```diff
@@ -10,32 +10,38 @@ internal class WireMockServerLifecycleHook(ILoggerFactory loggerFactory) : IDist
...
 
-        var wireMockServerResources = appModel.Resources
-            .OfType<WireMockServerResource>()
-            .ToArray();
-
-        foreach (var wireMockServerResource in wireMockServerResources)
+        _mappingTask = Task.Run(async () =>
```
StefH, line 20: maybe use SafeFireAndForget (https://github.com/TheCodeTraveler/AsyncAwaitBestPractices) or something similar to indicate that this is a long running task and exceptions do not matter or are only logged?

src/WireMock.Net.Aspire/WireMockHealthCheck.cs
```diff
@@ -0,0 +1,44 @@
...
+
+    /// <inheritdoc />
+    public async Task<HealthCheckResult> CheckHealthAsync(HealthCheckContext context, CancellationToken cancellationToken = default)
+    {
+        if (!await IsHealthyAsync(resource.AdminApi.Value, cancellationToken))
+        {
+            return HealthCheckResult.Unhealthy("WireMock.Net is not healthy");
+        }
+
+        if (resource.ApiMappingState == WireMockServerResource.MappingState.NotSubmitted)
```
StefH: What if I post some mappings.
And then clean all mappings?

Will the HealthStatus be not-healthy?

StefH (changes requested):
some comments / questions

StefH:
@Zguy can you take a look at code comments?

src/WireMock.Net.Aspire/WireMockServerLifecycleHook.cs
Zguy: Done

src/WireMock.Net.Aspire/WireMockServerResource.cs
Zguy: Done

src/WireMock.Net.Aspire/WireMockServerLifecycleHook.cs
Zguy: Done

src/WireMock.Net.Aspire/WireMockHealthCheck.cs
Zguy: It should still report healthy. The idea is that it represents that the intended mappings (or none) has been sent to the server.

src/WireMock.Net.Aspire/WireMockServerLifecycleHook.cs
Zguy: I don't think pulling in a third party library is worth it for such a small thing, but I could add a comment explaining the intention.

Any exceptions thrown in the task will be re-thrown in the `DisposeAsync` method.

StefH: It's ok.

PR #1423, opened 2026-02-09
Add WebSockets

StefH:


illunix:
Hey, can I already use it? Is it publish on pre release version?

StefH:
@illunix 
You should be able to use the latest preview02 from MyGet

PR #1436, opened 2026-03-31
Moving Scenario state change before global response delay is set

jayaraman-venkatesan:
Moved the change of scenario state change before global response delay so that subsequent retries arriving during a delay period see the updated state and match the correct next mapping instead of re-matching same old one.

## Submitter checklist

- [x] Recommended: Join [WireMock Slack](https://slack.wiremock.org/) to get any help in `#help-contributing` or a project-specific channel like `#wiremock-java`

### DIFF
--- a/src/WireMock.Net.Minimal/Owin/WireMockMiddleware.cs
+++ b/src/WireMock.Net.Minimal/Owin/WireMockMiddleware.cs
@@ -122,6 +122,14 @@ internal class WireMockMiddleware(
                 }
             }
 
+            // Transition scenario state immediately after matching, before any delay (global or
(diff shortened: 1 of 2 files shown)


jayaraman-venkatesan:
@StefH  The windows tests seems to be flaky, any help reruning it would be appreciated

jayaraman-venkatesan:
Thanks for re-running @StefH 
but it failed in different step, which again i suspect is because of parallelized tests

It failed at adminmapping test, but Admin routes never have a Scenario set — so the moved code is completely bypassed for any POST /__admin/mappings request. my change should not have effect on admin API routing.

Can you help looking into it or help re-running it once more? 
Thank you

Issue #1421, opened 2026-01-23
Deactivate mapping without deleting it


cezarypiatek:
Hi,
I checked the documentation but couldn’t find anything related to this. It would be very helpful to have the ability to deactivate a mapping without deleting it. This would allow for proper post-test analysis.

In my case, many tests dispose of mappings as soon as they’re no longer needed to avoid collisions. After the tests finish, it becomes difficult to investigate why certain requests weren’t matched, because the mapping definitions have already been removed.

Instead of adding another attribute to the mapping itself, it might be better to extend the admin endpoint for returning mapping with new attribute that will allow to include deleted entries. I think this would be the simplest way to achieve the desired outcome.

This improvement would definitely improve WireMockInspector usability.

StefH:
So actually, you would like to have a soft-delete option?
So posting a HTTP DELETE to the admin-endpoint would not delete it, but de-activate?

cezarypiatek:
Correct. After soft delete there will be no longer in effect but it would be possible to get those mapping via admin api for diagnostic purpose.

jayaraman-venkatesan:
I was doing some triage and found that first part of soft delete can be implemented. We store active mappings in a dict<guid, Imapping> we can add a isEnabled field to Mappings.cs in that to hide/enable the mapping. 
We just need to make sure when we call FindBestMatch() we have to add this new field filter to consider only active mappings. 

@StefH  are you working on this? If not can I pick this up to implement the above strategy?

jayaraman-venkatesan:
I have raised a pull request for this.
@StefH please help review it whenever free

PR #1437, opened 2026-04-02
Added feature to enable and disable mappings

jayaraman-venkatesan:
feat: add soft-disable support for mappings (#1421)

## Summary
- Introduces `IsEnabled` flag on mappings (defaults to `true`) that prevents a mapping from being matched without deleting it, preserving it for post-test inspection
- Adds `PUT /__admin/mappings/{guid}/enable` and `PUT /__admin/mappings/{guid}/disable` admin endpoints
- Adds `EnableMappingAsync` / `DisableMappingAsync` to `IWireMockAdminApi` typed client
- `IsEnabled: false` can be set in POST/PUT mapping body and round-trips through `GET /mappings` and static mapping files
- Disabled mappings appear in `GET /mappings` and `server.Mappings` — only request matching skips them

## Changes
- `IRespondWithAProvider` — new `WithIsEnabled(bool)` builder method (fluent, consistent with `WithProbability` / `WithTitle`)
- `RespondWithAProvider` — stores `_isEnabled`, applies it during `RespondWith()`
- `WireMockServer.ConvertMapping` — calls `WithIsEnabled(false)` when `mappingModel.IsEnabled == false`
- `MappingConverter.ToMappingModel` — serializes `IsEnabled` as `false` only when disabled (null = enabled, sparse/backward-compatible)
- `WireMockServer.Admin` — two new regex path matchers, two route registrations, `MappingEnable` / `MappingDisable` handlers, `TryParseGuidFromSecondToLastSegment` helper
- `MappingMatcherTests` — fixed `InitMappings` helper and exception test mock to set `IsEnabled = true` (required after `.Where(m => m.IsEnabled)` was added); added `WhenMappingIsDisabled_ShouldReturnNull` test
- `WireMockAdminApiTests.IsEnabled` — 4 integration tests covering create-disabled, disable endpoint, enable endpoint, GET serialization

## Backward compatibility
- `IsEnabled` defaults to `true` everywhere — existing mappings, JSON files, and API clients are unaffected
- `IsEnabled` field is omitted from JSON when `true` (null) and only written when `false`

## Submitter checklist

- [x] Recommended: Join [WireMock Slack](https://slack.wiremock.org/) to get any help in `#help-contributing` or a project-specific channel like `#wiremock-java`

### DIFF
--- a/src/WireMock.Net.Abstractions/Admin/Mappings/MappingModel.cs
+++ b/src/WireMock.Net.Abstractions/Admin/Mappings/MappingModel.cs
@@ -62,9 +62,9 @@ public class MappingModel
     public int? TimesInSameState { get; set; }
 
     /// <summary>
-    /// Value to determing if the mapping is active
+    /// Value to determine if the mapping is disabled. Defaults to <c>null</c> (not disabled).
     /// </summary>
-    public bool? IsEnabled { get; set; }
+    public bool? IsDisabled { get; set; }
 
     /// <summary>
     /// The request model.
--- a/src/WireMock.Net.Abstractions/Admin/Mappings/MappingModel.cs
+++ b/src/WireMock.Net.Abstractions/Admin/Mappings/MappingModel.cs
@@ -55,11 +55,16 @@ public class MappingModel
     /// In case the value is null state will not be changed.
     /// </summary>
     public string? SetStateTo { get; set; }
-    
+
     /// <summary>
     /// The number of times this match should be matched before the state will be changed to the specified one.
     /// </summary>
     public int? TimesInSameState { get; set; }
+
+    /// <summary>
+    /// Value to determing if the mapping is active
+    /// </summary>
+    public bool? IsEnabled { get; set; }
 
     /// <summary>
     /// The request model.
@@ -100,7 +105,7 @@ public class MappingModel
     /// </summary>
     public object? Data { get; set; }
 
-    /// <summary> 
+    /// <summary>
     /// The probability when this request should be matched. Value is between 0 and 1. [Optional]
     /// </summary>
     public double? Probability { get; set; }
--- a/src/WireMock.Net.Minimal/Owin/MappingMatcher.cs
+++ b/src/WireMock.Net.Minimal/Owin/MappingMatcher.cs
@@ -19,7 +19,7 @@ internal class MappingMatcher(IWireMockMiddlewareOptions options, IRandomizerDou
         var possibleMappings = new List<MappingMatcherResult>();
 
         var mappings = _options.Mappings.Values
-            .Where(m=>m.IsEnabled)
+            .Where(m => !m.IsDisabled)
             .Where(m => m.TimeSettings.IsValid())
             .Where(m => m.Probability is null || _randomizerDoubleBetween0And1.Generate() <= m.Probability)
             .ToArray();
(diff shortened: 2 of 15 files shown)


test/WireMock.Net.Tests/WebSockets/WebSocketIntegrationTests.cs
```diff
@@ -813,9 +813,9 @@ public async Task WithWebSocketProxy_Should_Proxy_Binary_Messages()
 
         var receivedData = await client.ReceiveAsBytesAsync(cancellationToken: _ct);
 
-        await Task.Delay(500, _ct);
-
-        // Assert
+        // Assert immediately — receivedData is already in memory, no delay needed.
```
StefH: This change looks good, however can you remove this change in this PR?

I'll make a new PR which I use to fix these failing tests.

src/WireMock.Net.Abstractions/Admin/Mappings/MappingModel.cs
```diff
@@ -55,12 +55,17 @@ public class MappingModel
...
+
     /// <summary>
     /// The number of times this match should be matched before the state will be changed to the specified one.
     /// </summary>
     public int? TimesInSameState { get; set; }
 
+    /// <summary>
+    /// Value to determing if the mapping is active
+    /// </summary>
+    public bool? IsEnabled { get; set; }
```
StefH: It's better to make the functionality the other way around.

So
```
public bool IsDisabled { get; set; }
```

In this way,it's backwards compatible.

src/WireMock.Net.Minimal/Owin/MappingMatcher.cs
```diff
@@ -19,6 +19,7 @@ internal class MappingMatcher(IWireMockMiddlewareOptions options, IRandomizerDou
         var possibleMappings = new List<MappingMatcherResult>();
 
         var mappings = _options.Mappings.Values
+            .Where(m=>m.IsEnabled)
```
StefH: Do a format-file, so that the code is alligned (extra spaces)

StefH (changes requested):

StefH:
@jayaraman-venkatesan 
Can you take a look at my comments?

src/WireMock.Net.Abstractions/Admin/Mappings/MappingModel.cs
jayaraman-venkatesan: Are you suggesting to change only on mappingmodel or in IMappings too?

StefH: Yes, rename all properties to `IsDisabled` and change the business logic.

jayaraman-venkatesan:
@StefH I have made all the suggested changes

StefH:
@jayaraman-venkatesan 
One thing I was thinking about.
Technically it's possible to disable an Admin mapping?

jayaraman-venkatesan:
@StefH 

The 2 APIs (`/enable` and `/disable` ) to trigger this action is guarded by IsAdminInterface check.
`WithIsDisabled(true)` builder is only used when constructing user mappings
I can see even the PUT api to update the mapping has the same check. 

I dont think there are any more interfaces that can trigger this action. But please let me know if had missed something. 

If needed, on top of this I can also add if `.Where(m => m.IsAdminInterface || !m.IsDisabled)` check in  `MappingMatcher.cs` to ignore the flag.

StefH:
Ok. It's fine

Issue #1469, opened 2026-06-01
WireMock.Net doesn't play well with Aspire 13.4.0


roycornelissen:
### Describe the bug
Just upgraded our Aspire dependencies to 13.4.0 using the `aspire update` CLI tool. Now the orchestration will not start up anymore, no Docker container is started and the Aspire dashboard hangs and never appears.

I am not sure if this is a regression in Aspire, but it happens only with WireMock so decided to report it here.

The last lines in my trace level logs are:

```txt
trce: Aspire.Hosting.ApplicationModel.ResourceNotificationService[0]
      Version: 2 Resource aspire-dashboard/aspire-dashboard-udmuvwrf update published: ResourceType = Executable, CreationTimeStamp = (null), State = { Text = (null), Style = (null) }, IsHidden = True, HeathStatus = (null), ResourceReady = False, ExitCode = (null), Urls = { https = https://localhost:15888/login?t=2ea8a8a52597d71346920a6508164dec http = http://localhost:18888/login?t=2ea8a8a52597d71346920a6508164dec otlp-grpc = https://localhost:21888 }, EnvironmentVariables = {  }, Properties = {  }, HealthReports = {  }, Commands = {  }
info: Aspire.Hosting.ApplicationModel.WireMockServerResource[0]
      Checking Health status from WireMock.Net
dbug: Aspire.Hosting.Dcp.dcp.NetworkReconciler[0]
      Network created   {"ContainerNetwork": "/aspire-container-network", "Reconciliation": 2, "NetworkName": "aspire-session-network-gcuqfmyv-MyApi"}
```

We are using WireMock.Net in our orchestration as follows:

```csharp
var wireMock = builder.AddWireMock("wiremock")
    .WithArgs("--verbose");
```

### Expected behavior:
WireMock container should startup properly and Dashboard should appear.

### Test to reproduce
- 1 Make new Aspire orchestration using latest Aspire 13.4.0
- 2 Add WireMock (using latest 2.8.0)
- 3 Start AppHost

Fails to start.

### Other related info
When I remove WireMock from the orchestration, all my other services are started properly... Papercut, Azure Storage emulator, Azure Service Bus emulator, et cetera. Only WireMock is blocking.

StefH:
Thanks for reporting. I will take a look.

StefH:
https://github.com/wiremock/WireMock.Net/pull/1470

roycornelissen:
[...]

PR #1471, opened 2026-06-02
Fix WireMock Aspire startup deadlock

roycornelissen:
Fixes a WireMock Aspire startup deadlock introduced after the Aspire 13.4 upgrade.

-  `AddWireMock()` registers a lifecycle subscriber that previously awaited WireMock health and admin mapping setup inside
- `ResourceEndpointsAllocatedEvent`. In Aspire 13.4 this can block the startup event pipeline before the WireMock container has finished starting.

The fix returns from the Aspire event immediately and runs the WireMock admin setup in a background task, logging any failures instead of blocking AppHost startup.

## References

#1469 

## Submitter checklist

- [ ] The PR request is well described and justified, including the body and the references
- [ ] The PR title represents the desired changelog entry
- [ ] The repository's code style is followed (see the contributing guide)
- [ ] Test coverage that demonstrates that the change works as expected

### DIFF
--- a/examples-Aspire/AspireApp1.AppHost/AspireApp1.AppHost.csproj
+++ b/examples-Aspire/AspireApp1.AppHost/AspireApp1.AppHost.csproj
@@ -1,6 +1,6 @@
 <Project Sdk="Microsoft.NET.Sdk">
 
-    <Sdk Name="Aspire.AppHost.Sdk" Version="13.1.0" />
+    <Sdk Name="Aspire.AppHost.Sdk" Version="13.4.0" />
 
     <PropertyGroup>
         <OutputType>Exe</OutputType>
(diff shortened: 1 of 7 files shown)


roycornelissen:
@StefH as mentioned in the issue, this is Codex's attempt at diagnosing the issue. I see your PR also touches the same class and seems much better informed. So don't hesitate to dismiss this one.

StefH:
@roycornelissen 
Thanks for thinking along, and I think my fix does fix the issue, and this PR is more as a work around and not guaranteed that all is started.

I'll dismiss this one and create a new nuget in a few minutes.

roycornelissen:
I agree and thanks for the quick fix!

Issue #1477, opened 2026-06-22
Urls in settings require a port


DavidJohnWilliams:
### Describe the bug
If you use the `Urls` setting and don't supply a port, when starting under ASP.Net Core you receive a timeout error.

The actual error is a NRE from this line: https://github.com/wiremock/WireMock.Net/blob/cbc28c235c711a47dad5d628b3172e159968159d/src/WireMock.Net.Minimal/Owin/AspNetCoreSelfHost.cs#L152

`PortUtils.TryExtract` has null annotations to say that the `scheme` out parameter will be null if `TryExtract` returns `false`, but there is a null-forgiving operator on the next line so you get a NRE.

### Expected behavior:
Either no port is allowed and the code above should not run if `TryExtract` returns `false`; or the docs should be updated to explain that port is required on entries in `Urls`.

### Test to reproduce
- 1 - Add a url to `Urls` without a port
- 2 - Start the server

### Other related info
Side note: The exception thrown from `Start` is a timeout rather than the actual exception that occurred which made debugging the issue more difficult. It would be good to solve that as well.

BTW, I am happy to create a PR for this, but I'm raising the issue first for discussion.

StefH:
@DavidJohnWilliams 
Thank you.

---

My suggestion:
- throw ArgumentException if TryExtract returns false here
- also throw for this code `PortUtils.TryExtract(addresses[i]` ?
- if possible ; add unit test

---

Side note: The exception thrown from Start is a timeout rather than the actual exception that occurred which made debugging the issue more difficult. It would be good to solve that as well.
--> make a new issue + PR for this one

DavidJohnWilliams:
@StefH I have gone through the code a bit more in detail and I can see why you are suggesting to throw an exception there.

A couple of questions:

1. Why does `PortUtils.TryExtract` use a regex instead of `Uri.TryCreate`?

    If we used `Uri.TryCreate` and a port was not supplied then it defaults to the correct port based on scheme (or -1 e.g. for grpc). This could allow more granular control over when to throw the exception if a port is not supplied.

2. Regardless of the above should we actually throw an exception in `HostUrlOptions.GetDetails` or `AspNetCoreSelfHost.SetHttpsAndUrls` as that is what is used to setup listening. And if the parsing fails there will be no listeners setup anyway.

3. If you want to keep port as mandatory, should docs be updated with valid URL format?

StefH:
[...]

PR #1478, opened 2026-06-26
Support default port for known schemas

DavidJohnWilliams:
- Update tests which expected urls without port to fail
- Add new tests for default ports

## References

- #1477

## Submitter checklist

### DIFF
--- a/test/WireMock.Net.Tests/Util/PortUtilsTests.cs
+++ b/test/WireMock.Net.Tests/Util/PortUtilsTests.cs
@@ -21,14 +21,14 @@ public class PortUtilsTests
         isGrpc.Should().BeFalse();
         scheme.Should().BeNull();
         host.Should().BeNull();
-        port.Should().Be(default(int));
+        port.Should().Be(default);
     }
 
     [Fact]
-    public void PortUtils_TryExtract_UrlIsMissingPort_Returns_False()
+    public void PortUtils_TryExtract_DefaultPort_UnknownScheme_Returns_False()
     {
         // Assign
-        var url = "http://0.0.0.0";
+        var url = "grpc://0.0.0.0";
 
         // Act
         var result = PortUtils.TryExtract(url, out var isHttps, out var isGrpc, out var scheme, out var host, out var port);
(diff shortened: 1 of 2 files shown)


StefH (approved):

test/WireMock.Net.Tests/Util/PortUtilsTests.cs
```diff
@@ -39,7 +39,43 @@ public void PortUtils_TryExtract_UrlIsMissingPort_Returns_False()
...
+        scheme.Should().Be("http");
+        host.Should().Be("0.0.0.0");
+        port.Should().Be(80);
+    }
+
+    [Fact]
+    public void PortUtils_TryExtract_DefaultPort_Https_Returns_True()
+    {
+        // Assign
+        var url = "https://0.0.0.0";
```
StefH, line 67: Can you check if you can also add a "real" unit test which starts wiremock at https;//0.0.0.0 or https://localhost with that port?

Like this:
https://github.com/wiremock/WireMock.Net/blob/master/test/WireMock.Net.Tests/WireMockServerTests.cs#L199

(it could be that that test runs locally correct, and maybe in CI-CD fails, maybe, we have to see)

StefH (changes requested):

DavidJohnWilliams:
@StefH I tried adding a test as you suggested and it fails I believe because the port is already in use.

I guess something to do with running as a hosted service under IIS vs running in a unit test context.

It's essentially calling the same code `WireMockServer.Start` but I'm guessing there is a lot going on under the hood that I don't know about.

StefH:
I guess you test locally in Windows?

---

If you create a simple console-app, and use that setup, does it also fail?

StefH:
@DavidJohnWilliams 
If you create a simple console-app, and use that setup, does it also fail?

---

And if this keeps failing, then I my assumption is that providing a port is mandatory and does make sense. So in that case, no change in the code is needed, only updating the docs is required. Agree?

DavidJohnWilliams:
Hi, sorry for the delay. I tried creating a console app. It actually fails whether I include a port or not.

For background we are hosting our instance of WireMock in IIS under "https://customdomain.local" which is configured with port 443 and a certificate.

In the console app, I tried:

- Custom URL no port -> Socket error
- Custom URL with port 443 -> Socket error
- Custom URL with random port -> Works fine

However when hosting under IIS both of the first two options work.

As I said before, I think if the docs are clear that a port is required, then I am happy with that. I mostly raised this because it was a breaking change for us. We had been running without a port specified, and didn't see anything in the changelog or docs about requiring one.

StefH:
Closing this PR in favor of documentation change.
https://github.com/wiremock/wiremock.org/pull/428

PR #1502, opened 2026-08-17
Fix #1501: Generate WhenStateIs/WillSetStateTo correctly in ToCSharpCode

### LINKED ISSUE #1501: Generated C# code for stateful (Scenario) mappings does not compile and drops the WhenStateIs condition
## Summary

The C# code generator exposed via `GET /__admin/mappings/code` produces incorrect code for any mapping that uses a Scenario/State.

For a mapping configured with `.WhenStateIs(...)` and/or `.WillSetStateTo(...)`, the generator emits:

```csharp
.WhenStateIs("<NextState>", <TimesInSameState>)
```

This is wrong in two ways:

1. It does not compile. There is no `WhenStateIs(string, int)` overload, only `WhenStateIs(string)` and `WhenStateIs(int)` exist. So the generated code cannot be pasted back into a project and built.
2. It silently drops the execution condition. The mapping's actual `WhenStateIs` value is never emitted at all, so the condition disappears from the generated code.

## Steps to reproduce

```csharp
var server = WireMockServer.StartWithAdminInterface();

// A stateful (Scenario) mapping:
//  - only eligible when the scenario "checkout" is in state "StepA"
//  - moves the scenario to state "StepB" after it has matched twice
server
    .Given(Request.Create().WithPath("/order").UsingGet())
    .InScenario("checkout")
    .WhenStateIs("StepA")
    .WillSetStateTo("StepB", 2)
    .RespondWith(Response.Create().WithSuccess());

// Generate the equivalent C# code for the mappings.
using var http = new HttpClient();
var code = await http.GetStringAsync($"{server.Url}/__admin/mappings/code");
Console.WriteLine(code);
```

The generated output contains:

```csharp
server
    .Given(Request.Create()
        .UsingMethod("GET")
        .WithPath(...)
    )
    .WithGuid("...")
    .InScenario("checkout")
    .WhenStateIs("StepB", 2)          // does not compile; wrong value; "StepA" is lost
    .RespondWith(Response.Create()
        .WithStatusCode(200)
    );
```

Trying to compile the generated snippet fails, e.g.:

```
error CS1501: No overload for method 'WhenStateIs' takes 2 arguments
```

## Expected behavior

The generated code should reconstruct the original mapping and compile as-is:

```csharp
    .InScenario("checkout")
    .WhenStateIs("StepA")
    .WillSetStateTo("StepB", 2)
```

## Actual behavior

- The generated code uses a non-existent `WhenStateIs(string, int)` overload, so it does not compile.
- The `NextState` value is emitted under `WhenStateIs`.
- The real `WhenStateIs` condition is omitted entirely.

## Notes

- Only C# code generation is affected; JSON serialization and round-trip of scenario state is correct.

Bafyn:
## Summary

`MappingConverter.ToCSharpCode` generated broken code for stateful (Scenario) mappings: it emitted `.WhenStateIs(NextState, TimesInSameState)`, which has no matching overload (does not compile), used `NextState` under `WhenStateIs`, and never emitted the actual `ExecutionConditionState`.

## References

Fixes #1501

## Submitter checklist

## Tests
* Added two `MappingConverterTests.ToCSharpCode` cases; 
* Corrected two `MappingBuilderTests.ToCSharpCode` verified snapshots.


StefH:
I `just` started also on this....

StefH:
@Bafyn 
Can you also update the Wildcardmatcher code to:

``` c#
public override string GetCSharpCodeArguments()
{
    return
        $"new {Name}" +
        $"(" +
        $"{MatchBehaviour.GetFullyQualifiedEnumValue()}, " +
        $"new AnyOf<string, StringPattern>[] {{ {MappingConverterUtils.ToCSharpCodeArguments(_patterns)} }}, " +
        $"{CSharpFormatter.ToCSharpBooleanLiteral(IgnoreCase)}, " +
        $"{MatchOperator.GetFullyQualifiedEnumValue()}" +
        $")";
}
```

Bafyn:
@StefH I realized you did when I noticed you assigned the issue to yourself 🙃
I can also look into this thing you mentioned https://github.com/wiremock/WireMock.Net/issues/1501#issuecomment-5318616837

**UPD**: yes, checking

src/WireMock.Net.Minimal/Serialization/MappingConverter.cs
```diff
@@ -167,9 +167,15 @@ public string ToCSharpCode(IMapping mapping, MappingConverterSettings? settings
         {
```
StefH, line 167: Change comment to:
```
 // Scenario, ExecutionConditionState, NextState and StateTimes
```

src/WireMock.Net.Minimal/Serialization/MappingConverter.cs
```diff
@@ -167,9 +167,15 @@ public string ToCSharpCode(IMapping mapping, MappingConverterSettings? settings
...
             sb.AppendLine($"    .InScenario({ToCSharpStringLiteral(mapping.Scenario)})");
         }
+        if (!string.IsNullOrEmpty(mapping.ExecutionConditionState))
+        {
+            sb.AppendLine($"    .WhenStateIs({ToCSharpStringLiteral(mapping.ExecutionConditionState)})");
+        }
         if (!string.IsNullOrEmpty(mapping.NextState))
         {
-            sb.AppendLine($"    .WhenStateIs({ToCSharpStringLiteral(mapping.NextState)}, {ToCSharpIntLiteral(mapping.TimesInSameState)})");
+            sb.AppendLine(mapping.TimesInSameState > 1
```
StefH: No need for the ternary statement.

In case the TimesInSameState is null, a null is already added.

test/WireMock.Net.Tests/Serialization/MappingConverterTests.ToCSharpCode.cs
```diff
@@ -97,6 +97,67 @@ public Task ToCSharpCode_With_Server_And_AddStartIsFalse()
...
+        // Verify
+        return Verify(code, VerifySettings);
+    }
+
+    private Mapping CreateScenarioMapping(string scenario, string? executionConditionState, string? nextState, int? timesInSameState)
+    {
+        var request = Request.Create().UsingGet().WithPath("/test_path");
+        var response = Response.Create().WithSuccess();
+
+        return new Mapping
```
StefH, line 240: Can you add the argument names here please?

Like:
``` c#
var mapping = new Mapping
(
    guid,
    _updatedAt,
    title: string.Empty,
    description: string.Empty,
    path: null,
    _settings,
    request,
    response,
    priority: 0,

    webhooks: null,
    useWebhooksFireAndForget: false,
    timeSettings: null,
    data: null
);
```

I forgot that also in the existing, can you fix that also?

StefH (changes requested):

src/WireMock.Net.Minimal/Serialization/MappingConverter.cs
Bafyn: it's there mainly to distinguish between 2 overloads: `(string state)` and `(string state, int? times = 1)`.

Still prefer to remove it and always use the 2-param one?

StefH: Just make it easy, always use 2 param

StefH (changes requested):

Bafyn:
FYI @StefH the other matchers had the same issues. Fixed them in scope of the latest update

src/WireMock.Net.Minimal/Matchers/ContentTypeMatcher.cs
```diff
@@ -83,7 +83,7 @@ public override string GetCSharpCodeArguments()
         return $"new {Name}" +
                $"(" +
                $"{MatchBehaviour.GetFullyQualifiedEnumValue()}, " +
-               $"{MappingConverterUtils.ToCSharpCodeArguments(_patterns)}, " +
+               $"new AnyOf<string, StringPattern>[] {{ {MappingConverterUtils.ToCSharpCodeArguments(_patterns)} }}, " +
```
StefH: I was thinking,

For the `MatchBehaviour` I used the GetFullyQualifiedEnumValue, for this AnyType, maybe it's better to also make sure to use the full type. So `new AnyOfTypes.AnyOf<string, StringPattern>[]` to be sure that the user is not required to import suing.

Can you fix that for all?

StefH (changes requested):

StefH:
Added comment.

And if you have fixed an issue, you can "resolve conversation" here in the github.

src/WireMock.Net.Minimal/Matchers/ContentTypeMatcher.cs
Bafyn: I think I had that idea too, but then rejected it for some reason. 
ok, can change it

StefH: yes pls change

Bafyn: the matchers were the only component that was missing full names, so applied the same change to them

StefH (approved):

PR #1504, opened 2026-08-28
Update SharpYaml package to version 2.5.1

### LINKED ISSUE #130: StackOverflowException with deeply nested flow sequences/mappings
## SharpYaml version

3.3.0 (latest)

## Environment

- Linux (WSL2, Ubuntu 22.04)
- .NET 10.0 (SDK 10.0.103 / Runtime 10.0.3)
- Also reproducible on any OS / .NET version

## Description

Deeply nested YAML flow sequences or mappings cause an unrecoverable `StackOverflowException` due to unbounded recursion in the parser and DOM loader. The mutual recursion between `YamlNode.ReadElement` ↔ `YamlSequence.Load` (and similarly `YamlMapping.Load`) has no depth limit, so sufficiently nested input exhausts the call stack.

`StackOverflowException` is particularly severe in .NET because it **cannot be caught** by `try`/`catch` — it terminates the process unconditionally. Any application that parses untrusted YAML input using SharpYaml is vulnerable to denial-of-service via a crafted document.

The threshold is approximately:
- ~21,000 nested flow sequences (`[[[...]]]`) for `YamlStream.Load`
- ~25,000 for `YamlSerializer.Deserialize`
- ~20,000 nested flow mappings (`{a: {a: ...}}`) for `YamlStream.Load`

These thresholds vary by platform and available stack size, and could be significantly lower in environments with smaller stacks (e.g., ASP.NET thread pool threads, which default to 1 MB).

## Steps to Reproduce

```csharp
using SharpYaml.Model;

// ~21,000 nested flow sequences — process terminates with StackOverflowException
int depth = 21000;
string input = new string('[', depth) + new string(']', depth);

// This kills the process — cannot be caught
var stream = YamlStream.Load(new StringReader(input));
```

Also reproducible via `YamlSerializer.Deserialize`:

```csharp
using SharpYaml;

int depth = 25000;
string input = new string('[', depth) + new string(']', depth);

_ = YamlSerializer.Deserialize<object>(input);
```

And with nested flow mappings:

```csharp
using SharpYaml.Model;
using System.Text;

int depth = 20000;
var sb = new StringBuilder();
for (int i = 0; i < depth; i++) sb.Append("{a: ");
sb.Append("{}");
for (int i = 0; i < depth; i++) sb.Append('}');

var stream = YamlStream.Load(new StringReader(sb.ToString()));
```

## Root Cause

The recursive descent has no depth counter or limit at any level:

1. **DOM layer**: `YamlNode.ReadElement` → `YamlSequence.Load` → `ReadElement` → ... (no depth check)
2. **Parser layer**: `ParseNode` → `ParseFlowSequenceEntry` → `ParseNode` → ... (no depth check)
3. **Scanner layer**: `flowLevel` tracks depth but is never checked against a maximum

## Comparison with Other Libraries

For reference, here's how other .NET parsing libraries handle deeply nested input:

| Library | Default Max Depth | Behavior on Overflow |
|---------|------------------|---------------------|
| System.Text.Json | **64** | Catchable `JsonException` |
| Newtonsoft.Json (v13+) | **64** | Catchable `JsonReaderException` |
| SharpYaml 3.3.0 | **None** | Unrecoverable `StackOverflowException` |

Both major .NET JSON libraries enforce a default depth limit of 64 with a clean, catchable exception. The YAML spec does not mandate a depth limit, but Java's SnakeYAML added one ([configurable, default varies](https://github.com/snakeyaml/snakeyaml/commit/fc300780da21f4bb92c148bc90257201220cf174)) after similar DoS concerns.

## Suggested Fix

The `Scanner` class already tracks nesting depth via `flowLevel` (incremented in `IncreaseFlowLevel`, called from `FetchFlowCollectionStart` when encountering `[` or `{`). It just never checks it against a maximum. Adding a limit there would protect all higher-level APIs (`YamlStream`, `EventReader`, `YamlSerializer`):

```csharp
// In Scanner.IncreaseFlowLevel() — add a check before incrementing
private void IncreaseFlowLevel()
{
    if (flowLevel >= MaxFlowLevel)
        throw new SemanticErrorException(/*...*/,
            $"Maximum nesting depth of {MaxFlowLevel} exceeded.");

    // existing code
    simpleKeys.Push(new SimpleKey());
    ++flowLevel;
}
```

A default of 64 would match the .NET ecosystem convention (System.Text.Json and Newtonsoft.Json both default to 64). This converts an unrecoverable process crash into a catchable `SemanticErrorException`, consistent with how SharpYaml already handles other malformed input.

## Impact

- **Severity**: High (DoS — process termination, not catchable)
- **Attack vector**: Any application parsing untrusted YAML (web APIs, config file processors, CI/CD pipelines)
- **Input size**: ~42 KB for flow sequences, ~100 KB for mappings — small enough to include in HTTP requests

Found by coverage-guided fuzzing with SharpFuzz + AFL++.

johnthcall:
## References

Upgrading Microsoft.OpenApi packages to a version which takes SharpYalm 2.1.5 with the following fix xoofx/SharpYaml#130

## Submitter checklist

### DIFF
--- a/src/WireMock.Net.OpenApiParser/WireMock.Net.OpenApiParser.csproj
+++ b/src/WireMock.Net.OpenApiParser/WireMock.Net.OpenApiParser.csproj
@@ -54,7 +54,7 @@
         <PackageReference Include="ILRepack.Lib.MSBuild.Task" Version="2.0.43" PrivateAssets="All" />
         <PackageReference Include="Microsoft.OpenApi" Version="3.7.0" PrivateAssets="All" />
         <PackageReference Include="Microsoft.OpenApi.YamlReader" Version="3.7.0" PrivateAssets="All" />
-        <PackageReference Include="SharpYaml" Version="2.1.4" />
+        <PackageReference Include="SharpYaml" Version="2.1.5" />
     </ItemGroup>
 
     <ItemGroup>


src/WireMock.Net.OpenApiParser/WireMock.Net.OpenApiParser.csproj
```diff
@@ -54,11 +54,11 @@
         <PackageReference Include="ILRepack.Lib.MSBuild.Task" Version="2.0.43" PrivateAssets="All" />
         <PackageReference Include="Microsoft.OpenApi" Version="3.7.0" PrivateAssets="All" />
         <PackageReference Include="Microsoft.OpenApi.YamlReader" Version="3.7.0" PrivateAssets="All" />
-        <PackageReference Include="SharpYaml" Version="2.1.4" />
+        <PackageReference Include="SharpYaml" Version="2.1.5" />
```
StefH: I see that there is also a new Microsoft.OpenApi / Microsoft.OpenApi.YamlReader which includes this version, can you upgrade that one instead and maybe remove SharpYaml dependency?

StefH (changes requested):

src/WireMock.Net.OpenApiParser/WireMock.Net.OpenApiParser.csproj
johnthcall: Upgraded to Microsoft.OpenApi and Microsoft.OpenApi.YamlReader to 3.10.2 and remove the SharpYaml dependency

johnthcall:
@StefH the following change was introduced in the latest Microsoft.OpenApi.YamlReader microsoft/OpenAPI.NET#3027 which added limits MaxAliasExpansionNodeCount. This can be overridden when creating OpenApiYamlReader in the constructor of WireMockOpenApiParser. WireMockOpenApiParserTests.FromText_UsingYaml_ShouldReturnMappings requires increasing the AliasExpansionNodeCount from 5k to over 100k. Should WireMockOpenApiParser take a nullable  OpenApiYamlReaderSettings or integer MaxAliasExpansionNodeCount to allow customers to increase this? Or should I revert the PR to just be the SharpYaml version increase?

StefH:
Yes please keep it to sharpyml.

StefH (approved):

StefH:
@johnthcall 

Why would the AliasExpansionNodeCount be need to increased from 5k to over 100k?

5000 sounds a lot to me ?

StefH:
It seems 1 million is needed for that example...

StefH:
https://github.com/StefH/RamlToOpenApiConverter/pull/30

StefH:
https://github.com/StefH/RamlToOpenApiConverter/pull/29

StefH:
https://github.com/wiremock/WireMock.Net/pull/1505

